PR #2, opened 2020-12-09
Add steps (server)

antalszava:
**Context**
This PR adds the server-side code of the PennyLane-Orquestra plugin. It is stand-alone from other parts of the plugin (and can be reviewed as such).

As part of Orquestra convention, workflows that are submitted to Orquestra must contain step definitions. A step is basically a piece of code to be executed and it can be made available by publishing it on a public repo on GitHub. Once done, it can be defined within workflows and Orquestra will pull the source to execute the given step.

**Changes**

Add the `run_circuit_and_get_expval` function inside of `steps` to define an Orquestra workflow step. As its name suggests, this function runs a circuit and computes the expectation value of operators.

Its inputs are:
* backend_specs (dict): the parsed Orquestra backend specifications
* circuit (str): the circuit represented as an OpenQASM 2.0 program
* operators (str): the operator in an ``openfermion.QubitOperator`` or ``openfermion.IsingOperator`` representation

**Further notes**

* The step definition is meant to integrate with [Orquestra interfaces](http://docs.orquestra.io/other-resources/interfaces/) (in specific with `QuantumBackend` and `QuantumSimulator`) and involves calling high-level functions. This is meant to preserve current and future modularity provided by Orquestra.
* `run_circuit_and_get_expval` includes applying the identity in cases when an empty circuit was provided or when a qubit is measured where no gates were applied. This is a temporary convenience to allow working with certain Orquestra backends.
* Certain viable features (e.g., supporting noise models with Qiskit devices, allowing sampling) are left as future additions.


.github/workflows/step_tests.yml
```diff
@@ -0,0 +1,57 @@
...
+
+jobs:
+  core-tests:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        config:
+          - {python-version: 3.7}
+          - {python-version: 3.8}
```
thisac, line 21: Is there a reason for 3.6 not being part of these tests (as they are in the other test file)?

.github/workflows/step_tests.yml
```diff
@@ -0,0 +1,57 @@
...
+        run: docker run --rm -d -p 5555:5555 rigetti/quilc -R
+
+      - name: Run Forest QVM
+        run: docker run --rm -d -p 5000:5000 rigetti/qvm -S
+
+      - name: Run tests
+        run: python -m pytest steps/test_* $COVERAGE_FLAGS
+
+      - name: Adjust coverage file for Codecov
+        run: bash <(sed -i 's/filename=\"/filename=\"pennylane_orquestra_steps\//g' coverage.xml)
```
thisac: Might not be needed as per [Josh's earlier comment](https://github.com/PennyLaneAI/pennylane-orquestra/pull/1/files?file-filters%5B%5D=.yml#r540381208). :slightly_smiling_face:

steps/expval.py
```diff
@@ -0,0 +1,157 @@
...
+    backend_specs: dict,
+    circuit: str,
+    operators: str,
+):
+    """Takes a circuit to obtain the expectation value of an operator on a
+    given backend.
+
+    All backend calls used in this function are defined as ``QuantumBackend``
+    and ``QuantumSimulator`` interface standard methods implemented are defined
+    in the ``z-quantum-core`` repository.
```
thisac: This sentence was a bit difficult to parse for me; it feels a bit wrong. :confused:

steps/expval.py
```diff
@@ -0,0 +1,157 @@
...
+
+    There are two computation modes: sampling and exact. Expectation values are
+    computed by post-processing samples when an Orquestra ``QuantumBackend`` is
+    used or the number of samples was specified for a ``QuantumSimulator``
+    backend. When the number of samples was not specified, ``QuantumSimulator``
```
thisac:
```suggestion
    backend. When the number of samples isn't specified, ``QuantumSimulator``
```

steps/expval.py
```diff
@@ -0,0 +1,157 @@
...
+    backends run in exact mode.
+
+    Args:
+        backend_specs (dict): the parsed Orquestra backend specifications
```
thisac: It seems like the `backend_specs` passed here are actually strings (although dict-like), which also explains why you need `json.loads`. Maybe `str` should be specified here as well as for the type hinting above, with an explanation that the string _must_ contain a dict (in a way)?

steps/expval.py
```diff
@@ -0,0 +1,157 @@
...
+
+    # 2. Create operators
+    ops = []
+    for op in operators:
+        if backend.n_samples is not None:
+            # Operator for Backend/Simulator in sampling mode
+            ops.append(IsingOperator(op))
+        else:
+            # Operator for Simulator exact mode
+            ops.append(QubitOperator(op))
```
thisac: Very minor; I feel like it's slightly clearer to check if sampled only once. Otherwise it feels like `backend.n_samples` would be able to change depending on the operator (at least that was my initial thought), unless I'm missing something. 

```suggestion
    if backend.n_samples is not None:
        # Operators for Backend/Simulator in sampling mode
        for op in operators:
            ops.append(IsingOperator(op))
    else:
        # Operators for Simulator exact mode
        for op in operators:
            ops.append(QubitOperator(op))
```

steps/expval.py
```diff
@@ -0,0 +1,157 @@
...
+        active_qubits.extend(instruction_qubits)
+
+    active_qubits = set(active_qubits)
+
+    # Get the qubits we'd like to measure
+    # Data for identities is not stored, need to account for empty terms
+    op_qubits = [term[0][0] for op in ops for term in op.terms if term]
+
+    need_to_activate = set(op_qubits) - active_qubits
+    if not need_to_activate == set():
```
thisac: Wouldn't it be enough to simply check `if need_to_activate:`? :thinking: 
```suggestion
    if need_to_activate:
```

steps/steps_requirements.txt
```diff
@@ -0,0 +1,10 @@
+qiskit==0.18.3
+qiskit-ibmq-provider==0.6.1
+pyquil==2.17.0
+numpy>=1.18.1
+git+https://github.com/zapatacomputing/qe-qiskit#dev
+git+https://github.com/zapatacomputing/qe-qulacs#dev
+git+https://github.com/zapatacomputing/qe-forest#dev
+git+https://github.com/zapatacomputing/z-quantum-core.git#dev
+git+https://github.com/zapatacomputing/qe-openfermion#dev
```
thisac, line 9: :dizzy_face:

thisac (commented):
Looking good @antalszava! Mainly had some minor comments. Awesome job with this! :muscle:

.github/workflows/step_tests.yml
antalszava: Yes, Orquestra doesn't support Python version <3.7.

steps/expval.py
antalszava: Fair point, thanks! Yeah, backend specs are usually represented as a json with Orquestra.

steps/expval.py
antalszava: Oh, good point! Rephrased it.

.github/workflows/step_tests.yml
antalszava: Thanks for catching that! Updated.

steps/expval.py
antalszava: Think that makes sense! Accepting it.

steps/expval.py
antalszava: Good catch, thanks!

antalszava:
Thanks @thisac for the comments! :slightly_smiling_face:

.github/workflows/step_tests.yml
josh146: The dream 😍

steps/expval.py
```diff
@@ -0,0 +1,159 @@
...
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+The PennyLane-Orquestra step for computing the expectation value of Hermitian
+operators given a quantum circuit and a quantum device.
```
josh146, line 17: Out of curiosity, do they not already have a step very similar to this in core Orquestra? This seems like a step that would be required for VQE etc

steps/expval.py
```diff
@@ -0,0 +1,159 @@
...
+from zquantum.core.utils import create_object, save_list
+
+
+def run_circuit_and_get_expval(
+    backend_specs: str,
+    circuit: str,
+    operators: str,
+):
+    """Takes a circuit to obtain the expectation value of an operator on a
+    given backend.
```
josh146:
```suggestion
    """Executes a circuit to obtain the expectation value of an operator on a
    given backend.
```

steps/expval.py
```diff
@@ -0,0 +1,159 @@
...
+        operators (str): the operator in an ``openfermion.QubitOperator``
+            or ``openfermion.IsingOperator`` representation
+    """
+    backend_specs = json.loads(backend_specs)
+    operators = json.loads(operators)
+
+    backend = create_object(backend_specs)
+
+    # 1. Parse circuit
+    qc = QuantumCircuit.from_qasm_str(circuit)
```
josh146, line 64: These comments are super useful!

steps/expval.py
```diff
@@ -0,0 +1,159 @@
...
+    # 2. Create operators
+    ops = []
+    if backend.n_samples is not None:
+        # Operators for Backend/Simulator in sampling mode
+        for op in operators:
+            ops.append(IsingOperator(op))
+    else:
+        # Operators for Simulator exact mode
+        for op in operators:
+            ops.append(QubitOperator(op))
```
josh146, line 75: Could you remind me why we using IsingOperators for sampling, but qubit operators for exact mode?

steps/expval.py
```diff
@@ -0,0 +1,159 @@
...
+
+    # 2.+1
+    # Activate the qubits that are measured but were not acted on
+    # By applying the identity
+    # Note: this is a temporary logic subject to be removed once supported by
+    # Orquestra
```
josh146, line 81: Does orquestra force you to measure all qubits? What happens if you don't?

steps/expval.py
```diff
@@ -0,0 +1,159 @@
...
+            val = np.sum(expectation_values.values)
+            results.append(val)
+    else:
+        for op in ops:
+            expectation_values = backend.get_exact_expectation_values(circuit, op)
+
+            val = np.sum(expectation_values.values)
+            results.append(val)
+
+    return results
```
josh146, line 158: Nice! This function was well written and easy to read.

steps/steps_requirements.txt
josh146: 😵😵😵

steps/test_expval.py
```diff
@@ -0,0 +1,173 @@
...
+
+    @pytest.mark.parametrize("op", ['["[Z0]"]', '["[Z1]"]', '["[Z2]"]', '["[]"]'])
+    def test_only_measure_circuit(self, op, backend_specs, monkeypatch):
+        """Tests that the correct result in obtained for a circuit that only
+        contains measurements."""
+        lst = []
+
+        only_measure_qasm = 'OPENQASM 2.0;\ninclude "qelib1.inc";\nqreg q[3];\ncreg c[3];\n'
+
+        monkeypatch.setattr(expval, "save_list", lambda val, name: lst.append(val))
```
josh146, line 64: I couldn't quite work out what this monkeypatching was doing?

josh146: Is this simply a way of extracting the expectation values early?

steps/test_expval.py
```diff
@@ -0,0 +1,173 @@
...
+        contains a Hadamard gate."""
+        lst = []
+
+        hadamard_qasm = 'OPENQASM 2.0;\ninclude "qelib1.inc";\nqreg q[2];\ncreg c[2];\nh q[0];\n'
+        op = '["[Z0]"]'
+
+        monkeypatch.setattr(expval, "save_list", lambda val, name: lst.append(val))
+
+        expval.run_circuit_and_get_expval(backend_specs, hadamard_qasm, op)
+        assert math.isclose(lst[0][0], 0.0, abs_tol=analytic_tol)
```
josh146, line 82: Maybe it would be also worth testing a case where the exact expectation value is non-zero

steps/test_expval.py
```diff
@@ -0,0 +1,173 @@
...
+        contains a Hadamard gate."""
+        lst = []
+
+        hadamard_qasm = 'OPENQASM 2.0;\ninclude "qelib1.inc";\nqreg q[2];\ncreg c[2];\nh q[0];\n'
+        op = '["[Z0]"]'
+
+        monkeypatch.setattr(expval, "save_list", lambda val, name: lst.append(val))
+
+        expval.run_circuit_and_get_expval(backend_specs, hadamard_qasm, op)
+        assert math.isclose(lst[0][0], 0.0, abs_tol=tol)
```
josh146, line 140: Same here re: a non-zero expectation value

steps/test_expval.py
```diff
@@ -0,0 +1,173 @@
...
+
+        lst = []
+
+        monkeypatch.setattr(expval, "save_list", lambda val, name: lst.append(val))
+        expval.run_circuit_and_get_expval(dev.backend_specs, qasm_circuit, ops)
+        expected = np.array(
+            [np.sin(theta) * np.sin(phi) + np.cos(theta), np.cos(theta) * np.cos(phi) + np.sin(phi)]
+        ) / np.sqrt(2)
+
+        assert np.allclose(lst[0], expected, atol=analytic_tol)
```
josh146: 😻

josh146: Oh I just realised it is marked as xfail 😆 

I would suggest adding the reason to the `xfail` decorator, e.g.,

```python
@pytest.mark.xfail(reason="no way of currently testing this")
```

How come this test is expected to fail? Is the functionality coming in a later PR?

josh146 (commented):
@antalszava I need to give you kudos for splitting these PRs up in such a way that they are a joy to review

steps/expval.py
antalszava: Not quite like this. The step that they have for the VQE tutorial incorporates other Orquestra interfaces (e.g., optimizers, etc.) and accepts ansatze that are pre-defined rather than dynamic (e.g., `SingletUCCSDAnsatz`):
* [VQE tutorial](https://github.com/zapatacomputing/z-quantum-vqe/blob/master/examples/hydrogen-vqe.yaml)
* [VQE step used to optimize](https://github.com/zapatacomputing/z-quantum-optimizers/blob/88de904dd5ee2ce58d51d0a6a59ad8afe31aa28b/steps/optimizers.py#L14)

Internally, these functions are used, however, to compute expectation values.

josh146: Got it!

steps/expval.py
antalszava: This is how Orquestra computes expectation values internally:
```python
        # We require operator to be IsingOperator because measurements are always performed in the Z basis,
        # so we need the operator to be Ising (containing only Z terms).
        # A general Qubit Operator could have X or Y terms which don’t get directly measured.
        if isinstance(ising_operator, IsingOperator) == False:
            raise Exception("Input operator is not openfermion.IsingOperator")
```
For the [`get_expectation_values` function that is called on samples](https://github.com/zapatacomputing/z-quantum-core/blob/master/src/python/zquantum/core/measurement.py).

steps/expval.py
antalszava: An  `ValueError: Invalid number of qubits specified` is raised for certain simulators (e.g., Qiskit). There is an `if` statement internally which checks the number of qubits and it keeps track of the number of active qubits. Therefore an error is raised if e.g., 1 qubit was activated out of 3, but the 2nd qubit is measured. Therefore we need to activate all the qubits where we'd like to have measurements.

See snippet:
```python
from openfermion import QubitOperator
import json
from zquantum.core.utils import create_object
from qiskit import QuantumCircuit
from zquantum.core.circuit import Circuit
backend_specs = '{"module_name": "qeqiskit.simulator", "function_name": "QiskitSimulator", "device_name": "statevector_simulator"}'
op = "1 [Z2]"
circuit = '''OPENQASM 2.0;
      include "qelib1.inc";
      qreg q[3];
      creg c[3];
      x q[0];
        '''
backend_specs = json.loads(backend_specs)
backend = create_object(backend_specs)
qc = QuantumCircuit.from_qasm_str(circuit)
circuit = Circuit(qc)
op = QubitOperator(op)
expectation_values = backend.get_exact_expectation_values(circuit, op)
```

steps/test_expval.py
antalszava: With Orquestra, at the end of each step we'd serialize the result by dumping it into a `json` file. This is done using the internal `save_list` function in the `run_circuit_and_get_expval` step.

The monkeypatching disables calling this function and instead passes it into a local list so that the value can be checked within the test.

steps/test_expval.py
antalszava: Yes, it uses the `"orquestra.qulacs"` device the logic for which is not yet added. Will remove the `xfail` mark whenever that gets added.

steps/expval.py
josh146: Huh. I find that a super strange restriction; why not just check if the IsingOperator _or_ QubitOperator consists of only `Z`.

steps/expval.py
josh146: Got it!

steps/test_expval.py
josh146: Maybe worth adding a comment, super helpful thanks Antal!

steps/test_expval.py
antalszava: Thanks! Added.

steps/test_expval.py
antalszava: Thanks! Added.

steps/test_expval.py
antalszava: Sure thing! Added.

steps/expval.py
antalszava: Hmm, I vaguely remember as if it was required due to how openfermion is being used at times. There's a convenience function used at times to convert between types of openfermion operators:
https://github.com/zapatacomputing/qe-openfermion/blob/6d52df917727dc71883b5672ce8015fd917a6d98/src/python/qeopenfermion/_utils.py#L350

steps/expval.py
```diff
@@ -0,0 +1,159 @@
...
+    # Note: this is a temporary logic subject to be removed once supported by
+    # Orquestra
+
+    # Get the active qubits of the circuit
+    active_qubits = []
+    for instr in qc.data:
+        instruction_qubits = [qubit.index for qubit in instr[1]]
+        active_qubits.extend(instruction_qubits)
+
+    active_qubits = set(active_qubits)
```
agran2018: maybe one can save this line and using ``set`` below ?
```suggestion
```

steps/expval.py
```diff
@@ -0,0 +1,159 @@
...
+
+    # Get the qubits we'd like to measure
+    # Data for identities is not stored, need to account for empty terms
+    op_qubits = [term[0][0] for op in ops for term in op.terms if term]
+
+    need_to_activate = set(op_qubits) - active_qubits
```
agran2018: here
```suggestion
    need_to_activate = set(op_qubits) - set(active_qubits)
```

steps/test_expval.py
```diff
@@ -72,11 +74,36 @@ def test_run_circuit_and_get_expval_hadamard(self, backend_specs, monkeypatch):
...
+        # Instead of creating a json at the end of the step, just append it to
+        # a local list
+        monkeypatch.setattr(expval, "save_list", lambda val, name: lst.append(val))
+
+        expval.run_circuit_and_get_expval(backend_specs, circ_qasm, op)
+
+        res = np.array(lst[0])
+        assert np.allclose(
+            res, np.array([np.cos(theta), np.cos(theta) * np.cos(phi)]), atol=analytic_tol
+        )
```
josh146, line 105: 💪

steps/expval.py
```diff
@@ -0,0 +1,159 @@
...
+    # If there are still no instructions, apply identity to the first qubit
+    # Can happen for an empty circuit when measuring the identity operator
+    if not qc.data:
+        qc.id(qc.qubits[0])
+
+    # Convert to zquantum.core.circuit.Circuit
+    circuit = Circuit(qc)
+
+    # 3. Expval
+    results = _get_expval(backend, circuit, ops)
```
josh146, line 109: I'm assuming the number of shots is passed implicitly as `backend.n_samples`?

josh146 (approved):
Thanks for addressing all my suggestions @antalszava!

steps/expval.py
```diff
@@ -0,0 +1,159 @@
...
+
+    if backend.n_samples is not None:
+        measurements = backend.run_circuit_and_measure(circuit)
+
+        # Iterating through the operators specified e.g., [IsingOperator("[Z0]
+        # + [Z1]"), IsingOperator("[Z1]")] to post-process the measurements
+        # outcomes
+        for op in ops:
+            expectation_values = measurements.get_expectation_values(op)
+            expectation_values = expectation_values_to_real(expectation_values)
```
agran2018, line 144: just out of curiosity: if ``op`` is hermitian, why is the output of ``measurements.get_expectation_values(op)`` not real?

agran2018 (commented):
Thanks @antalszava, left just very very minor comments.  will go through it later to continue digesting things. I am basically learning about this plugin

steps/expval.py
antalszava: Good question! Went with how expectation values are computed internally for Orquestra backends (like for the [Qulacs simulator](https://github.com/zapatacomputing/qe-qulacs/blob/76c42cbf1d065e8fb94bc165ceb2b84c7327daff/src/python/qequlacs/simulator.py#L45)). This sort of operation is usually a safeguard to make sure that the type of the value returned is real (a complex value could sometimes cause problems.

steps/expval.py
antalszava: Yes, the number of samples is stored with
```python
    backend_specs = json.loads(backend_specs)
    backend = create_object(backend_specs)
```

steps/expval.py
```diff
@@ -0,0 +1,158 @@
...
+    defined in the ``z-quantum-core`` repository.
+
+    There are two computation modes: sampling and exact. Expectation values are
+    computed by post-processing samples when an Orquestra ``QuantumBackend`` is
+    used or the number of samples was specified for a ``QuantumSimulator``
+    backend. When the number of samples isn't specified, ``QuantumSimulator``
+    backends run in exact mode.
+
+    Args:
+        backend_specs (str): the Orquestra backend specification in a json
```
thisac, line 52: :partying_face:

steps/expval.py
```diff
@@ -0,0 +1,158 @@
...
+    else:
+        # Operators for Simulator exact mode
+        for op in operators:
+            ops.append(QubitOperator(op))
+
+    # 2.+1
+    # Activate the qubits that are measured but were not acted on
+    # By applying the identity
+    # Note: this is a temporary logic subject to be removed once supported by
+    # Orquestra
```
thisac: Looked a bit off having the `B` in `By` capitalized. (minor)
```suggestion
    # Activate the qubits that are measured but were not acted on
    # by applying the identity
    # Note: this is a temporary logic subject to be removed once supported by
    # Orquestra
```

steps/test_expval.py
```diff
@@ -0,0 +1,229 @@
...
+            qml.RY(theta, wires=[0])
+            qml.RY(phi, wires=[1])
+            qml.CNOT(wires=[0, 1])
+            return qml.expval(qml.Hadamard(wires=0)), qml.expval(qml.Hadamard(wires=1))
+
+        qnode = qml.QNode(circuit, dev)
+        qnode._construct([], {})
+        qasm_circuit = dev.serialize_circuit(qnode.circuit)
+        ops, _ = dev.process_observables(qnode.circuit.observables)
+        ops = json.dumps(ops)
```
thisac, line 184: Am I understanding this correctly; it takes the QNode and transforms it into a qasm-circuit string (via `dev.serialize_circuit`), and also extracts the observables (as a list?) and creates a string out of it stored in `ops`?

steps/test_expval.py
```diff
@@ -0,0 +1,229 @@
...
+@pytest.fixture
+def token():
+    """Get the IBMQX test token."""
+    t = os.getenv("IBMQX_TOKEN_TEST", None)
+
+    if t is None:
+        pytest.skip("Skipping test, no IBMQ token available")
+
+    yield t
+    IBMQ.disable_account()
```
thisac, line 208: Nice! :fire:

thisac: Does this mean that this test is skipped in the CI (or is there somehow a token there)?

thisac (approved):
:100:

steps/test_expval.py
antalszava: Yes. `ops` will contain the serialized operations. This test is marked as `xfail` for now, as it uses the `OrquestraDevice` capabilities (to be adjusted in #4).

steps/test_expval.py
antalszava: That's a good catch! It's skipped for now but will look into configuring it for the CI.

antalszava:
Thanks for all the comments and reviews! :confetti_ball: Merging it for now, but can revisit if need be.

PR #3, opened 2020-12-10
Add cli actions

antalszava:
**Context**

At the moment, submitting workflows as jobs to Orquestra can be done only via the Quantum Engine command line tool.

**Changes**

This PR adds functions that make `qe` CLI calls for submitting workflows and getting their details & results.

These functions are specifically tailored to the Orquestra API v1.0.0 version and can be sensitive in their functioning. The aim was to keep dependencies on the responses sent by Orquestra to a minimum.

**Benefits**

The Orquestra platform is available from Python scripts.

**Drawbacks**
Change in the responses sent by Orquestra can lead to breaking changes.

### DIFF
--- a/pennylane_orquestra/cli_actions.py
+++ b/pennylane_orquestra/cli_actions.py
@@ -42,7 +42,8 @@ def qe_get(workflow_id, option="workflow"):
             Examples include ``workflow`` and ``workflowresult``.
 
     Returns:
-        str: response message of the CLI call
+        list: the list of strings representing the response message of the CLI
+        call
     """
     process = subprocess.Popen(
         ["qe", "get", option, str(workflow_id)], stdout=subprocess.PIPE, universal_newlines=True
@@ -102,7 +103,8 @@ def workflow_details(workflow_id):
             retrieved
 
     Returns:
-        str: response message of the CLI call
+        list: the list of strings representing the response message of the CLI
+        call
     """
     return qe_get(workflow_id, option="workflow")
 
@@ -119,7 +121,8 @@ def workflow_results(workflow_id):
             Examples include ``workflow`` and ``workflowresult``.
 
     Returns:
-        str: response message of the CLI call
+        list: the list of strings representing the response message of the CLI
+        call
     """
     return qe_get(workflow_id, option="workflowresult")
 
--- a/pennylane_orquestra/cli_actions.py
+++ b/pennylane_orquestra/cli_actions.py
@@ -218,6 +218,8 @@ def loop_until_finished(workflow_id, timeout=300):
     # Setting filename=None will treat the file as temporary and it will be
     # removed
     file_tmp = urllib.request.urlretrieve(location, filename=None)[0]
+
+    # Data is retrieved as a tarball, need to extract the json file within
     if tarfile.is_tarfile(file_tmp):
         tar = tarfile.open(file_tmp, "r:gz")
         tar.extractall()
--- a/pennylane_orquestra/cli_actions.py
+++ b/pennylane_orquestra/cli_actions.py
@@ -1,3 +1,17 @@
+# Copyright 2020 Xanadu Quantum Technologies Inc.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 """
 This module contains utilities and auxiliary functions for using the Orquestra
 Quantum Engine command line interface (CLI).
--- a/pennylane_orquestra/cli_actions.py
+++ b/pennylane_orquestra/cli_actions.py
@@ -16,12 +16,12 @@
 This module contains utilities and auxiliary functions for using the Orquestra
 Quantum Engine command line interface (CLI).
 """
-import subprocess
-import time
-import os
-import urllib.request
 import json
+import os
+import subprocess
 import tarfile
+import time
+import urllib.request
 
 import yaml
 from appdirs import user_data_dir
--- a/pennylane_orquestra/cli_actions.py
+++ b/pennylane_orquestra/cli_actions.py
@@ -215,7 +215,7 @@ def loop_until_finished(workflow_id, timeout=300):
             continue
 
     # 3. Obtain the data from the URL
-    # Seting filename=None will treat the file as temporary and it will be
+    # Setting filename=None will treat the file as temporary and it will be
     # removed
     file_tmp = urllib.request.urlretrieve(location, filename=None)[0]
     if tarfile.is_tarfile(file_tmp):
--- a/pennylane_orquestra/cli_actions.py
+++ b/pennylane_orquestra/cli_actions.py
@@ -222,7 +222,7 @@ def loop_until_finished(workflow_id, timeout=300):
     # removed
     file_tmp = urllib.request.urlretrieve(location, filename=None)[0]
 
-    # Data is retrieved as a tarball, need to extract the json file within
+    # Data is retrieved as a tarfile, need to extract the json file within
     if tarfile.is_tarfile(file_tmp):
         tar = tarfile.open(file_tmp, "r:gz")
         tar.extractall()
@@ -230,5 +230,7 @@ def loop_until_finished(workflow_id, timeout=300):
 
         with open("workflow_result.json") as json_file:
             data = json.load(json_file)
+    else:
+        raise ValueError(f"Unexpected datatype for the retrieved result: not a tarfile.")
 
     return data
--- a/pennylane_orquestra/cli_actions.py
+++ b/pennylane_orquestra/cli_actions.py
@@ -131,7 +131,7 @@ def write_workflow_file(filename, workflow):
     """Write a workflow file given the name of the file.
 
     This function will create a YAML file with the workflow content. The file
-    is placed into a user specific data folder specified by using
+    is placed into a user-specific data folder specified by using
     ``appdirs.user_data_dir``.
 
     Args:
--- a/pennylane_orquestra/cli_actions.py
+++ b/pennylane_orquestra/cli_actions.py
@@ -170,6 +170,11 @@ def loop_until_finished(workflow_id, timeout=300):
 
     Returns:
         dict: the resulting dictionary parsed from a json file
+
+    Raises:
+        TimeoutError: if no response was obtained after the timeout
+        ValueError: if the workflow execution failed or if the result was not
+        returned as a tarfile
     """
     start = time.time()
     query_results = True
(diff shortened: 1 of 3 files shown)


pennylane_orquestra/cli_actions.py
```diff
@@ -0,0 +1,231 @@
...
+    """Function for submitting a workflow via a CLI call.
+
+    Handling submission response messages was based on using Orquestra API
+    v1.0.0.
+
+    Args:
+        filepath (str): the filepath for the workflow file
+
+    Keyword Args:
+        keep_file=False (bool): whether or not to keep or delete the workflow
```
co9olguy, line 64: How come you put in this option? And why is it `False` by default?

I'm guessing it's because these workflow files are just emphemeral, i.e., they are generated by PL-Orquestra plugin, but don't represent a workflow the user would actually want to repeat?

pennylane_orquestra/cli_actions.py
```diff
@@ -0,0 +1,231 @@
...
+
+    details = "".join(res)
+    if "Success" not in details:
+        raise ValueError(res)
+
+    if not keep_file:
+        os.remove(filepath)
+
+    unexpected_resp_msg = "Received an unexpected response after submitting workflow."
+    if isinstance(res, list):
```
co9olguy, line 86: Is receiving a list signalling a bad outcome?

pennylane_orquestra/cli_actions.py
```diff
@@ -0,0 +1,231 @@
...
+        option (str): The option specified for the ``qe get`` CLI call.
+            Examples include ``workflow`` and ``workflowresult``.
+
+    Returns:
+        str: response message of the CLI call
+    """
+    process = subprocess.Popen(
+        ["qe", "get", option, str(workflow_id)], stdout=subprocess.PIPE, universal_newlines=True
+    )
+    return process.stdout.readlines()
```
co9olguy, line 51: Just wanted to check that this `readlines()` returns a full string and not a list of strings?

pennylane_orquestra/cli_actions.py
```diff
@@ -0,0 +1,231 @@
...
+
+def workflow_details(workflow_id):
+    """Function for getting workflow details via a CLI call.
+
+    Args:
+        workflow_id (str): the workflow id for which information will be
+            retrieved
+
+    Returns:
+        str: response message of the CLI call
```
co9olguy: Depending on the answer to my questoin above, you might also need to update the return type here

pennylane_orquestra/cli_actions.py
```diff
@@ -0,0 +1,231 @@
...
+    Args:
+        workflow_id (str): the workflow id for which information will be
+            retrieved
+
+    Kwargs:
+        option (str): The option specified for the ``qe get`` CLI call.
+            Examples include ``workflow`` and ``workflowresult``.
+
+    Returns:
+        str: response message of the CLI call
```
co9olguy: same

pennylane_orquestra/cli_actions.py
```diff
@@ -0,0 +1,231 @@
...
+    """
+    return qe_get(workflow_id, option="workflowresult")
+
+
+def write_workflow_file(filename, workflow):
+    """Write a workflow file given the name of the file.
+
+    This function will create a YAML file with the workflow content. The file
+    is placed into a user specific data folder specified by using
```
co9olguy:
```suggestion
    is placed into a user-specific data folder specified by using
```

pennylane_orquestra/cli_actions.py
```diff
@@ -0,0 +1,231 @@
...
+        except urllib.error.URLError:
+            continue
+
+    # 3. Obtain the data from the URL
+    # Setting filename=None will treat the file as temporary and it will be
+    # removed
+    file_tmp = urllib.request.urlretrieve(location, filename=None)[0]
+
+    # Data is retrieved as a tarball, need to extract the json file within
+    if tarfile.is_tarfile(file_tmp):
```
co9olguy, line 231: Might be a bit dangerous to have an `if` here but no else (raise exception?)

pennylane_orquestra/cli_actions.py
antalszava: The thought here was not to clutter the disk of the user without them knowing that. `qe_submit` is going to be called by the devices in `PennyLane-Orquestra` and a workflow file will be created and submitted under the hood.

> I'm guessing it's because these workflow files are just emphemeral, i.e., they are generated by PL-Orquestra plugin, but don't represent a workflow the user would actually want to repeat?

Exactly. The `keep_file` option will be available to be toggled by passing the same `kwarg` to a device. That way a user can look at the exact workflow file if they wanted to (or re-submit it in the future independent of using PennyLane-Orquestra).

pennylane_orquestra/cli_actions.py
antalszava: Nope, that's okay. Receiving a list would actually be expected here. We'd also expect that we find the workflow ID in the second element of the list, so an `IndexError` is raised if that's not the case.

pennylane_orquestra/cli_actions.py
antalszava: Great catch, thanks! Updated at all occurrences.

tests/test_cli_actions.py
```diff
@@ -0,0 +1,211 @@
...
+from pennylane_orquestra.cli_actions import (loop_until_finished, qe_submit,
+                                             write_workflow_file)
+
+
+class TestCLIFunctions:
+    """Test functions for CLI actions work as expected."""
+
+    def test_submit_raises_no_success(self, monkeypatch):
+        """Test that the qe_submit method raises an error if not a successful
+        message was received."""
```
co9olguy:
```suggestion
        """Test that the qe_submit method raises an error if a success
        message was not received."""
```

tests/test_cli_actions.py
```diff
@@ -0,0 +1,211 @@
...
+                result_message = ["Some message2", "Some location"]
+
+                m.setattr(
+                    pennylane_orquestra.cli_actions, "workflow_results", lambda *args: result_message
+                )
+                m.setattr(urllib.request, "urlopen", lambda arg: arg)
+
+                # test_file is a json file instead of a tarfile
+                m.setattr(urllib.request, "urlretrieve", lambda *args, **kwargs: (test_file,))
+                loop_until_finished("Some ID", timeout=1)
```
co9olguy, line 191: Where is the `assert` this test?

co9olguy: according to the docstring, I expected to find a `pytest.raises` or similar

co9olguy: Nevermind, I see it now, it's just very far from the actual function being tested :smile:

co9olguy: Could you move the `pytest.raises` to be closer to the last 1-2 lines of the test?

co9olguy (approved):
Thanks @antalszava, solid code :muscle: 

Note: I did not test the CLI directly

co9olguy:
(I ran the tests though, and they pass)

tests/test_cli_actions.py
```diff
@@ -0,0 +1,203 @@
...
+            result_message = ["Some message2", "Some location"]
+
+            m.setattr(
+                pennylane_orquestra.cli_actions, "workflow_results", lambda *args: result_message
+            )
+            m.setattr(urllib.request, "urlopen", lambda arg: arg)
+            m.setattr(urllib.request, "urlretrieve", lambda *args, **kwargs: (test_tar,))
+            assert loop_until_finished("Some ID", timeout=1) == decoded_data
+
+    # Expected to fail if qe is unavailable
```
co9olguy: When exactly do we expect this to fail? When the `qe` CLI is not available, or when the remote service is not accessible?

antalszava: Good question! It's when `qe` is unavailable. Added this so that the CI can pass.

co9olguy: Weird. I ran it locally (where `qe` is available) and it still `xfail`d

tests/test_cli_actions.py
antalszava: For sure! :slightly_smiling_face:  Thanks for the comment, updated the test.

tests/test_cli_actions.py
antalszava: Hmm that sounds odd :thinking: Did it have `X` (upper case) or `x` (lower case)? For me locally it's `X` (passed but marked as `xfail`). Not logged in via `qe`.

pennylane_orquestra/cli_actions.py
```diff
@@ -0,0 +1,236 @@
...
+"""
+This module contains utilities and auxiliary functions for using the Orquestra
+Quantum Engine command line interface (CLI).
+"""
+import subprocess
+import time
+import os
+import urllib.request
+import json
+import tarfile
```
josh146: Best to order these alphabetically!

pennylane_orquestra/cli_actions.py
```diff
@@ -0,0 +1,236 @@
...
+
+import yaml
+from appdirs import user_data_dir
```
josh146, line 27: Same here :)

pennylane_orquestra/cli_actions.py
```diff
@@ -0,0 +1,236 @@
...
+    """
+    return qe_get(workflow_id, option="workflowresult")
+
+
+def write_workflow_file(filename, workflow):
+    """Write a workflow file given the name of the file.
+
+    This function will create a YAML file with the workflow content. The file
+    is placed into a user-specific data folder specified by using
+    ``appdirs.user_data_dir``.
```
josh146, line 135: Out of curiosity, what is the reason for writing to the user data directory, as opposed to allowing the directory to be specified solely through the `filename` argument?

pennylane_orquestra/cli_actions.py
```diff
@@ -0,0 +1,236 @@
...
+    start = time.time()
+    query_results = True
+    tries = 0
+    while query_results:
+        tries += 1
+
+        # Check if we've exceeded the timeout time, otherwise loop further
+        if time.time() - start > timeout:
+            current_status = workflow_details(workflow_id)
+            raise TimeoutError(
```
josh146, line 188: I didn't know about timeout error! nice

josh146 (approved):
Looks 💯 Antal! Very clear, concise, and easy to read and follow. My favourite type of code review!

pennylane_orquestra/cli_actions.py
antalszava: Thanks! Ran `isort` on the file.

pennylane_orquestra/cli_actions.py
antalszava: This function is going to be called by the device, thought it's good to place any files preserved into a well-defined directory that makes sense (instead of e.g., `cwd`).

pennylane_orquestra/cli_actions.py
antalszava: Oh was not documented, good point! Added to docstring

pennylane_orquestra/cli_actions.py
antalszava: Hmm, this seems to be the output of `isort`, fine to keep it as is?

antalszava:
Thanks @co9olguy @josh146 for the comments! Addressed them. Will merge, but can modify if a change is needed.

PR #4, opened 2020-12-14
Add devices base and utils

antalszava:
*(Note to reviewers: the size of the PR is due to a) adding the base device b) adding several device boilerplates and c) adding utilities which are to some extent code-reviewed already.)*

**Context**
For submitting workflows to Orquestra, a PennyLane device needs to send serialized data about:
* the backend to be used
* the circuit(s) to be run
* the operator(s) for which to obtain expectation values for

To address these needs, the approach taken is to have a base device (`OrquestraDevice`) that can take care of serialization and have further devices (`QeForestDevice`, `QeQiskitDevice`, etc.) that contain details needed for Orquestra.

A PennyLane user can simply instantiate a PennyLane device and the correct `backend_specs` will be specified for the Orquestra step:

```python
dev = qml.device("orquestra.qiskit", backend="statevector_simulator", wires=1, analytic=True)
```
is serialized to
```python
'{"module_name": "qeqiskit.simulator", "function_name": "QiskitSimulator", "device_name": "statevector_simulator"}'
```

**Changes**
* The base `OrquestraDevice` device
* Further devices (like `QeForestDevice`, `QeQiskitDevice`, etc.) including details needed as input to the workflows (`qe_component`, `qe_module_name`, etc.)
* Serialization of the backend specification, a circuit and operators as part of the `OrquestraDevice` device
* `utils.py`: `_terms_to_qubit_operator_string` and `_process_wires` functions adapted from PennyLane-Qchem

*What is not included?*
* `device.execute` & tests
* `device.batch_execute` & tests
* end-to-end tests

*Note*: devices will not send a workflow when computing the expectation value of the `Identity`, but instead use the theoretical value.

**Benefits**
* By slightly changing the code used in PennyLane-Qchem, the `openfermion` and PennyLane-Qchem dependencies (and sub dependencies) can be omitted as a direct dependency to the PennyLane-Orquestra plugin.

**Drawbacks**
* Code duplication for two functions.


pennylane_orquestra/utils.py
```diff
@@ -0,0 +1,195 @@
...
+"""This module contains utility functions for the PennyLane-Orquestra plugin.
+
+The original implementation of the ``_process_wires`` and
+``_terms_to_qubit_operator`` functions is a part of the PennyLane-QChem
+library.
+"""
+from pennylane.wires import Wires
+
+
+def _process_wires(wires, n_wires=None):
```
antalszava, line 23: This function is identical to the [one in PennyLane-Qchem](https://github.com/PennyLaneAI/pennylane/blob/88be620a04a802380d07dbdd84acb48a8f245fe6/qchem/pennylane_qchem/qchem/structure.py#L32). It was added separately to PennyLane-Orquestra to minimize the number of dependencies of the plugin.

pennylane_orquestra/utils.py
```diff
@@ -0,0 +1,195 @@
...
+    if len(wires) != n_wires:
+        # check length consistency when all checking and cleaning are done.
+        raise ValueError(
+            "Length of `wires` ({}) does not match `n_wires` ({})".format(len(wires), n_wires)
+        )
+
+    return wires
+
+
+def _terms_to_qubit_operator_string(coeffs, ops, wires=None):
```
antalszava, line 119: This function is an adaptation of [`_terms_to_qubit_operator` in PennyLane-Qchem](https://github.com/PennyLaneAI/pennylane/blob/88be620a04a802380d07dbdd84acb48a8f245fe6/qchem/pennylane_qchem/qchem/structure.py#L552). Instead of creating an `openfermion.QubitOperator` internally, it creates the string representation of the operator.

It was added separately to PennyLane-Orquestra to minimize the number of dependencies of the plugin.

pennylane_orquestra/ibmq_device.py
```diff
@@ -0,0 +1,57 @@
...
+
+    short_name = "orquestra.ibmq"
+
+    qe_component = "qe-qiskit"
+    qe_module_name = "qeqiskit.backend"
+    qe_function_name = "QiskitBackend"
+
+    def __init__(self, wires, shots=1024, backend="ibmq_qasm_simulator", **kwargs):
+
+        self._token = os.getenv("IBMQX_TOKEN") or kwargs.get("ibmqx_token", None)
```
thisac: This would mean that the environment variable would take precedence over the kwarg, right? Shouldn't it be the other way around; if a token is passed as a kwarg, then that one is used (even if there's an environmental var token)?
```suggestion
        self._token = kwargs.get("ibmqx_token", None) or os.getenv("IBMQX_TOKEN")
```

pennylane_orquestra/ibmq_device.py
```diff
@@ -0,0 +1,57 @@
...
+
+        if self._token is None:
+            raise ValueError(
+                "Please pass a valid IBMQX token to the device using the 'ibmqx_token' argument."
+            )
+
+        if "analytic" in kwargs and kwargs["analytic"]:
```
thisac:
```suggestion
        if kwargs.get("analytic", None):
```

pennylane_orquestra/ibmq_device.py
```diff
@@ -0,0 +1,57 @@
...
+            # Raise a warning if the analytic attribute was set to True
+            warnings.warn(
+                "The {self.short_name} device cannot be used in "
+                "analytic mode. Results are based on sampling."
+            )
+
+        kwargs["analytic"] = False
```
thisac: Is this in any way preferred to simply raising an error when analytic is `True`? 

It might be a bit confusing if a user passes `analytic=True`, gets this warning, and then it works (since analytic is set to `False`). I would prefer either including `Setting analytic to False` in the warning, or simply raising an error instead.

pennylane_orquestra/qiskit_device.py
```diff
@@ -0,0 +1,41 @@
...
+    def __init__(self, wires, shots=1024, backend="qasm_simulator", **kwargs):
+        if backend == "qasm_simulator":
+            if "analytic" in kwargs and kwargs["analytic"]:
+                # Raise a warning if the analytic attribute was set to True
+                warnings.warn(
+                    "The qasm_simulator backend device cannot be used in "
+                    "analytic mode. Results are based on sampling."
+                )
+
+            kwargs["analytic"] = False
```
thisac: Same comment as above!

pennylane_orquestra/qiskit_device.py
```diff
@@ -0,0 +1,41 @@
...
+
+    short_name = "orquestra.qiskit"
+
+    qe_component = "qe-qiskit"
+    qe_module_name = "qeqiskit.simulator"
+    qe_function_name = "QiskitSimulator"
+
+    def __init__(self, wires, shots=1024, backend="qasm_simulator", **kwargs):
+        if backend == "qasm_simulator":
+            if "analytic" in kwargs and kwargs["analytic"]:
```
thisac:
```suggestion
            if kwargs.get("analytic", None):
```

pennylane_orquestra/orquestra_device.py
```diff
@@ -0,0 +1,378 @@
...
+        self._batch_size = kwargs.get("batch_size", 10)
+        self._keep_files = kwargs.get("keep_files", False)
+        self._resources = kwargs.get("resources", None)
+        self._timeout = kwargs.get("timeout", 300)
+        self._latest_id = None
+        self._filenames = []
+        self._backend_specs = None
+
+    def apply(self, operations, **kwargs):
+        pass
```
thisac, line 129: Is this never used, or should this be added to later? :thinking:

pennylane_orquestra/orquestra_device.py
```diff
@@ -0,0 +1,378 @@
...
+            dict: the backend specifications represented as a dictionary
+        """
+        backend_specs = {}
+        backend_specs["module_name"] = self.qe_module_name
+        backend_specs["function_name"] = self.qe_function_name
+
+        if self.backend is not None:
+            # Only backends that have multiple backend need to specify one
+            # E.g., qe-qiskit
+            backend_specs["device_name"] = self.backend
```
thisac: This comment was slightly confusing (e.g. "backends with backends"); should it be `Only devices with multiple backends...`?. I'm assuming this is something that is always set, for all devices, but is only necessary for those with multiple possible backends?

pennylane_orquestra/orquestra_device.py
```diff
@@ -0,0 +1,378 @@
...
+        "S",
+        "SWAP",
+        "SX",
+        "T",
+        "Toffoli",
+    }
+
+    observables = {"PauliX", "PauliY", "PauliZ", "Identity", "Hadamard"}
+
+    def __init__(self, wires, shots=1000, analytic=True, **kwargs):
```
thisac: Is there a reason for 1000 shots being default? Wouldn't perhaps simply using 1 shot make more sense? :thinking:

pennylane_orquestra/orquestra_device.py
```diff
@@ -0,0 +1,378 @@
...
+    @abc.abstractmethod
+    def qe_function_name(self):
+        """Device specific Orquestra function name used in the backend
+        specification."""
+
+    @property
+    @abc.abstractmethod
+    def qe_component(self):
+        """Device specific Orquestra component name used in the backend
+        specification."""
```
thisac, line 217: Are these to be added to the devices later? Right now simply nothing happens if called; would it be better to raise an `NotImplementedError`?

pennylane_orquestra/orquestra_device.py
```diff
@@ -0,0 +1,378 @@
...
+        Args:
+            circuit (~.CircuitGraph): circuit to serialize
+
+        Returns:
+            str: OpenQASM 2.0 representation of the circuit without any
+            measurement instructions
+        """
+        qasm_str = circuit.to_openqasm(rotations=not self.analytic)
+
+        qasm_without_measurements = re.sub("measure.*?;\n", "", qasm_str)
```
thisac: Nice, nice! :fire: 

Just a thought on this: could there ever be statements after the measure statement on the same line? E.g. `measure bla bla; gate bla;`. In that case I think this would work instead (greedily checking for an extra `\n` and also removing any whitespace before the following statement on the same line):

```suggestion
        qasm_without_measurements = re.sub("measure.*?;\n?\s*", "", qasm_str)
```

pennylane_orquestra/orquestra_device.py
```diff
@@ -0,0 +1,378 @@
...
+                # Otherwise keep track of the indices and use the theoreticaly
+                # value as a result later
+                identity_indices.append(idx)
+
+        return ops, identity_indices
+
+    def serialize_operator(self, observable):
+        """
+        Serialize the observable specified for the circuit as an OpenFermion
+        operator.
```
thisac:
```suggestion
        """Serialize the observable specified for the circuit as an
        OpenFermion operator.
```

pennylane_orquestra/orquestra_device.py
```diff
@@ -0,0 +1,378 @@
...
+            wires (Wires): the wires the observable of the quantum function
+                acts on
+
+        Returns:
+            str: the ``openfermion.IsingOperator`` string representation
+        """
+        op_wires_but_last = [f"Z{w} " for w in wires[:-1]]
+        op_last_wire = f"Z{wires[-1]}"
+        op_str = "".join(["[", *op_wires_but_last, op_last_wire, "]"])
+        return op_str
```
thisac: Why do the last wire separately? Wouldn't just `""join([f"Z{w} " for w in wires])` work equally well? :thought_balloon:

pennylane_orquestra/orquestra_device.py
```diff
@@ -0,0 +1,378 @@
...
+        accepted_obs = {"PauliX", "PauliY", "PauliZ", "Identity"}
+
+        if isinstance(observable, Tensor):
+            need_decomposition = any(o.name not in accepted_obs for o in observable.obs)
+        else:
+            need_decomposition = observable.name not in accepted_obs
+
+        if need_decomposition:
+
+            original_observable = observable
```
thisac:
```suggestion
        if need_decomposition:
            original_observable = observable
```

thisac: Just to be sure; is the observable fully copied here, or could it be referencing?

tests/test_orquestra_device.py
```diff
@@ -0,0 +1,288 @@
...
+    (qml.PauliY(wires=[0]), "1 [Y0]"),
+    (qml.PauliZ(wires=[0]), "1 [Z0]"),
+    # Need decomposition
+    (qml.Hadamard(wires=[0]), "0.7071067811865475 [X0] + 0.7071067811865475 [Z0]"),
+    (qml.Hermitian(mx, wires=[0, 1]), "2.5 [] + -0.5 [Z1] + -1.0 [Z0]"),
+    (qml.Identity(wires=[0]) @ qml.Identity(wires=[1]), "1 []"),
+    (qml.PauliX(wires=[0]) @ qml.Identity(wires=[1]), "1 [X0]"),
+    (qml.PauliY(wires=[0]) @ qml.Identity(wires=[1]), "1 [Y0]"),
+    (qml.PauliZ(wires=[0]) @ qml.Identity(wires=[1]), "1 [Z0]"),
+    (qml.Hermitian(mx, wires=[0, 1]) @ qml.Identity(wires=[2]), "2.5 [] + -0.5 [Z1] + -1.0 [Z0]"),
```
thisac, line 167: :+1:

thisac (commented):
Looking great @antalszava. Had a few comments, suggestions, but mainly minor things/questions. :slightly_smiling_face: :rocket:

pennylane_orquestra/ibmq_device.py
antalszava: Good catch! Will update the Qiskit plugin too :smile: Thanks!

pennylane_orquestra/ibmq_device.py
antalszava: That's a good point! Overall the two cases seem to be pretty similar. One case where having a warning could be beneficial if this is part of a larger script and this device is used at a later stage. In such a case an error would disrupt the entire computation. So would rather keep it as is, updating the message sounds good!

pennylane_orquestra/orquestra_device.py
antalszava: Never used, but abstract in `QubitDevice`, so it needs to be redefined.

pennylane_orquestra/orquestra_device.py
antalszava: Makes sense, updated! Thanks! For devices that take no backend info, Orquestra doesn't need any in the `backend_specs`.

pennylane_orquestra/orquestra_device.py
antalszava: We usually have `1000` or `1024` as default for plugins.

pennylane_orquestra/orquestra_device.py
antalszava: These are defined by each device as class attributes. How do you mean? For me, it's raising a `TypeError` on the base class:
```python
In [6]: from pennylane_orquestra import OrquestraDevice                         

In [7]: dev = OrquestraDevice()                                                 
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-7-8fdf24cab504> in <module>
----> 1 dev = OrquestraDevice()

TypeError: Can't instantiate abstract class OrquestraDevice with abstract methods qe_component, qe_function_name, qe_module_name

```

pennylane_orquestra/orquestra_device.py
antalszava: Think at the moment no, as PennyLane doesn't allow applying gates after the measurements, but could come in handy. Thanks!

pennylane_orquestra/orquestra_device.py
thisac: Ooh, I missed that these were defined as class attributes. Now it makes sense. :smile:

pennylane_orquestra/orquestra_device.py
antalszava: No space is added after the last wire. Adding a comment for clarification.

pennylane_orquestra/orquestra_device.py
antalszava: `original_observable` is introduced only for readability, `observable` is not mutated (`decompose_hamiltonian` returns completely new observables, note that only `original_observable.matrix` is passed).

antalszava:
Hi @thisac, thanks so much for the comments! :100: Addressed them, they were really helpful!

pennylane_orquestra/orquestra_device.py
thisac: Oh, I see. :+1:

pennylane_orquestra/forest_device.py
```diff
@@ -0,0 +1,39 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+The Forest device class for PennyLane-Orquestra.
+"""
+from pennylane_orquestra.orquestra_device import OrquestraDevice
+
+
+class QeForestDevice(OrquestraDevice):
+    """Orquestra device"""
```
josh146: How come the docstring is so short? It doesn't specify details regarding the Forest nature of the device

josh146:
```suggestion
    """Orquestra Forest device"""
```

pennylane_orquestra/ibmq_device.py
```diff
@@ -0,0 +1,58 @@
...
+The IBMQ device class for PennyLane-Orquestra.
+"""
+import os
+import warnings
+
+from pennylane_orquestra.orquestra_device import OrquestraDevice
+
+
+class QeIBMQDevice(OrquestraDevice):
+    """The Orquestra IBMQ device."""
```
josh146: Similar question here - no details regarding the class args/kwargs?

pennylane_orquestra/ibmq_device.py
```diff
@@ -0,0 +1,58 @@
...
+
+        if self._token is None:
+            raise ValueError(
+                "Please pass a valid IBMQX token to the device using the 'ibmqx_token' argument."
+            )
+
+        if kwargs.get("analytic", None):
+            # Raise a warning if the analytic attribute was set to True
+            warnings.warn(
+                "The {self.short_name} device cannot be used in analytic "
```
josh146:
```suggestion
                f"The {self.short_name} device cannot be used in analytic "
```

pennylane_orquestra/forest_device.py
```diff
@@ -0,0 +1,39 @@
...
+
+    short_name = "orquestra.forest"
+
+    qe_component = "qe-forest"
+    qe_module_name = "qeforest.simulator"
+    qe_function_name = "ForestSimulator"
```
josh146, line 45: What's the purpose of these class attributes?

pennylane_orquestra/orquestra_device.py
```diff
@@ -0,0 +1,378 @@
...
+
+    The workflow files generated are placed into a user specific data folder
+    specified by the output of ``appdirs.user_data_dir("pennylane-orquestra",
+    "Xanadu")``. By default, such files are removed (see
+    ``keep_files`` keyword argument). After each device execution,
+    filenames for the generated workflows are stored in the ``filenames``
+    attribute.
+
+    Computing the expectation value of the identity operator does not involve a
+    workflow submission (hence no files are created).
```
josh146, line 54: :+1:

pennylane_orquestra/orquestra_device.py
```diff
@@ -0,0 +1,378 @@
...
+            number of samples set by the argument ``shots`` are used to estimate
+            these quantities.
+
+    Keyword Args:
+        backend=None (str): the Orquestra backend device to use for the
+            specific Orquestra backend, if applicable
+        batch_size=10 (int): the size of each circuit batch when using the
+            :meth:`~.batch_execute` method to send multiple workflows
+        keep_files=False (bool): Whether or not the workflow files
+            generated during the circuit execution should be kept or deleted.
```
josh146: If this is true, are new workflow files generated with unique names to avoid overwriting?

pennylane_orquestra/orquestra_device.py
```diff
@@ -0,0 +1,378 @@
...
+        resources=None (dict): An option for Orquestra, specifies the resources
+            provisioned for the clusters running each workflow step.
+        timeout=300 (int): The time until a job should timeout after getting no
+            response from Orquestra (in seconds).
```
josh146:
```suggestion
        timeout=300 (int): The maximum time until a job will timeout after getting no
            response from Orquestra (in seconds).
```

pennylane_orquestra/orquestra_device.py
```diff
@@ -0,0 +1,378 @@
...
+    """
+
+    name = "Orquestra base device for PennyLane"
+    short_name = "orquestra.base"
+    pennylane_requires = ">=0.11.0"
```
josh146:
```suggestion
    pennylane_requires = ">=0.13.0"
```

pennylane_orquestra/orquestra_device.py
josh146: I note that this differs from `orquestra.forest` above, which has `shots=1024` as default 🤔

josh146: @antalszava btw the convention of shots, by default, being 1000 or 1024 is completely arbitrary, I think I just picked that off my head when writing one of the plugins originally 😆 So if there is ever a good reason to choose a different value, feel free to change. Sometimes I think we should have gone higher by default.

pennylane_orquestra/orquestra_device.py
```diff
@@ -0,0 +1,378 @@
...
+        "RZ",
+        "Rot",
+        "S",
+        "SWAP",
+        "SX",
+        "T",
+        "Toffoli",
+    }
+
+    observables = {"PauliX", "PauliY", "PauliZ", "Identity", "Hadamard"}
```
josh146, line 114: I don't think this has to be specified (it is inherited from QubitDevice)

pennylane_orquestra/orquestra_device.py
josh146: Does it? 🤔 I thought it only needed to be defined if `OrquestraDevice` was _instantiated_, which should never be the case because it itself is also abstract

pennylane_orquestra/orquestra_device.py
```diff
@@ -0,0 +1,378 @@
...
+
+    def apply(self, operations, **kwargs):
+        pass
+
+    @classmethod
+    def capabilities(cls):
+        capabilities = super().capabilities().copy()
+        capabilities.update(
+            model="qubit",
+            supports_inverse_operations=False,
```
josh146, line 136: It doesn't support inverse operations?

pennylane_orquestra/orquestra_device.py
```diff
@@ -0,0 +1,378 @@
...
+            supports_analytic_computation=True,
+            returns_probs=False,
```
josh146, line 138: :(

pennylane_orquestra/orquestra_device.py
```diff
@@ -0,0 +1,378 @@
...
+            str: the backend specifications represented as a string
+        """
+        if self._backend_specs is None:
+            backend_specs_dict = self.create_backend_specs()
+            self._backend_specs = json.dumps(backend_specs_dict)
+
+        return self._backend_specs
+
+    def create_backend_specs(self):
+        """Create the backend specifications based on the device options.
```
josh146: It might be good to provide some more detail here!

pennylane_orquestra/orquestra_device.py
```diff
@@ -0,0 +1,378 @@
...
+    @abc.abstractmethod
+    def qe_function_name(self):
+        """Device specific Orquestra function name used in the backend
+        specification."""
+
+    @property
+    @abc.abstractmethod
+    def qe_component(self):
+        """Device specific Orquestra component name used in the backend
+        specification."""
```
josh146, line 217: Oh, this explains my question above!

pennylane_orquestra/orquestra_device.py
```diff
@@ -0,0 +1,378 @@
...
+        happens. Instead, the index of the observable is saved.
+
+        Args:
+            observables (list): a list of observables to process
+
+        Returns:
+            tuple:
+
+                * the serialized non-identity operators
+                * the indices of the identity operators
```
josh146, line 252: make sure to check that this renders correctly in the docs once ready :) The indentation might need to be adjusted

pennylane_orquestra/orquestra_device.py
```diff
@@ -0,0 +1,378 @@
...
+                identity_indices.append(idx)
+
+        return ops, identity_indices
+
+    def serialize_operator(self, observable):
+        """Serialize the observable specified for the circuit as an
+        OpenFermion operator.
+
+        Args:
+            observable (~.Observable): the observable to get the operator
```
josh146:
```suggestion
            observable (pennylane.operation.Observable): the observable to get the operator
```

pennylane_orquestra/orquestra_device.py
```diff
@@ -0,0 +1,378 @@
...
+
+        **Example**
+
+        >>> dev = QeQiskitDevice(wires=2)
+        >>> wires = [0, 1, 2]
+        >>> op_str = dev.pauliz_operator_string(wires)
+        >>> print(op_str)
+        [Z0 Z1 Z2]
+        >>> print(openfermion.IsingOperator(op_str))
+        1.0 [Z0 Z1 Z2]
```
josh146, line 304: nice example!

pennylane_orquestra/orquestra_device.py
```diff
@@ -0,0 +1,378 @@
...
+            op_str = self.pauliz_operator_string(wires)
+        else:
+            op_str = self.qubit_operator_string(observable)
+
+        return op_str
+
+    @staticmethod
+    def pauliz_operator_string(wires):
+        """Creates an OpenFermion operator string based on the related wires
+        that can be passed when creating an ``openfermion.IsingOperator``.
```
josh146:
```suggestion
        """Creates an OpenFermion operator string based on the measured wires
        that can be passed when creating an ``openfermion.IsingOperator``.
```
?

pennylane_orquestra/orquestra_device.py
```diff
@@ -0,0 +1,378 @@
...
+        op_wires_but_last = [f"Z{w} " for w in wires[:-1]]
+
+        # No space after the last wire
+        op_last_wire = f"Z{wires[-1]}"
+        op_str = "".join(["[", *op_wires_but_last, op_last_wire, "]"])
+        return op_str
+
+    def qubit_operator_string(self, observable):
+        """Creates an OpenFermion operator string from an observable that can
+        be passed when creating an ``openfermion.QubitOperator``.
```
josh146: I think this docstring (and the ones above) can be made more concise, e.g.,

> Serializes a PennyLane observable to a string compatible with the ``openfermion.QubitOperator`` class.

pennylane_orquestra/orquestra_device.py
```diff
@@ -0,0 +1,378 @@
...
+            need_decomposition = any(o.name not in accepted_obs for o in observable.obs)
+        else:
+            need_decomposition = observable.name not in accepted_obs
+
+        if need_decomposition:
+            original_observable = observable
+
+            # Decompose the matrix of the observable
+            # This removes information about the wire labels used and
+            # consecutive integer wires are used
```
josh146, line 357: @antalszava is this a bug in the `decompose_hamiltonian` function that we need to fix in PL core?

pennylane_orquestra/qiskit_device.py
```diff
@@ -0,0 +1,42 @@
...
+    qe_component = "qe-qiskit"
+    qe_module_name = "qeqiskit.simulator"
+    qe_function_name = "QiskitSimulator"
+
+    def __init__(self, wires, shots=1024, backend="qasm_simulator", **kwargs):
+        if backend == "qasm_simulator":
+            if kwargs.get("analytic", None):
+                # Raise a warning if the analytic attribute was set to True
+                warnings.warn(
+                    "The {self.short_name} device cannot be used in analytic "
```
josh146:
```suggestion
                    f"The {self.short_name} device cannot be used in analytic "
```

pennylane_orquestra/qulacs_device.py
```diff
@@ -0,0 +1,30 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+The Qulacs device class for PennyLane-Orquestra.
+"""
+from pennylane_orquestra.orquestra_device import OrquestraDevice
+
+
+class QeQulacsDevice(OrquestraDevice):
+    """Orquestra device"""
```
josh146:
```suggestion
    """The Orquestra qulacs device"""
```

pennylane_orquestra/qulacs_device.py
```diff
@@ -0,0 +1,30 @@
...
+
+    short_name = "orquestra.qulacs"
+
+    qe_component = "qe-qulacs"
+    qe_module_name = "qequlacs.simulator"
+    qe_function_name = "QulacsSimulator"
+
+    def __init__(self, wires, shots=1024, **kwargs):
+        super().__init__(wires, shots=shots, **kwargs)
```
josh146: does this need to be overwritten? It looks like nothing is being modified here

pennylane_orquestra/utils.py
```diff
@@ -0,0 +1,195 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""This module contains utility functions for the PennyLane-Orquestra plugin.
+
+The original implementation of the ``_process_wires`` and
+``_terms_to_qubit_operator`` functions is a part of the PennyLane-QChem
+library.
```
josh146, line 18: maybe good to add a link to the GitHub location of the function?

tests/test_orquestra_device.py
```diff
@@ -0,0 +1,293 @@
...
+    test_batch_res1,
+    test_batch_res2,
+    resources_default,
+    MockPopen,
+)
+
+qiskit_analytic_specs = '{"module_name": "qeqiskit.simulator", "function_name": "QiskitSimulator", "device_name": "statevector_simulator"}'
+qiskit_sampler_specs = '{"module_name": "qeqiskit.simulator", "function_name": "QiskitSimulator", "device_name": "statevector_simulator", "n_samples": 1000}'
+forest_noisy_specs = '{"module_name": "qeforest.simulator", "function_name": "ForestSimulator", "device_name": "3q-noisy-qvm", "n_samples": 10000}'
+ibmq_specs = '{"module_name": "qeqiskit.backend", "function_name": "QiskitBackend", "device_name": "ibmq_qasm_simulator", "n_samples": 1000, "api_token": "Some token"}'
```
josh146, line 36: Perhaps the fact that 10000 shots are needed here is a sign that we should increase the default number of shots on the devices? Although I _think_ Qiskit only supports max 8192 shots

tests/test_orquestra_device.py
```diff
@@ -0,0 +1,293 @@
...
+
+        qasm = dev.serialize_circuit(qnode.circuit)
+        expected = 'OPENQASM 2.0;\ninclude "qelib1.inc";\nqreg q[1];\ncreg c[1];\nh q[0];\n'
+        assert qasm == expected
+
+
+mx = np.diag(np.array([1, 2, 3, 4]))
+
+obs_serialize = [
+    # Don't decomposition
```
josh146:
```suggestion
    # Don't decompose
```

tests/test_orquestra_device.py
```diff
@@ -0,0 +1,293 @@
...
+]
+
+# More advanced examples for testing the correct wires
+obs_decomposed_wires_check = [
+    (qml.Hadamard(0), '0.7071067811865475 [X0] + 0.7071067811865475 [Z0]'),
+    (qml.Hadamard(2), '0.7071067811865475 [X2] + 0.7071067811865475 [Z2]'),
+    (qml.Hadamard(2) @ qml.Hadamard(1), '0.4999999999999999 [X2 X1] + 0.4999999999999999 [X2 Z1] + 0.4999999999999999 [Z2 X1] + 0.4999999999999999 [Z2 Z1]'),
+    (qml.Hermitian((qml.PauliY(1) @ qml.PauliX(0) @ qml.PauliZ(2)).matrix, wires=[1,0,2]), '1.0 [Y1 X0 Z2]'),
+    (qml.PauliY(2) @ qml.Hermitian((qml.PauliX(1)).matrix, wires=[1]), '1.0 [Y2 X1]')
+]
```
josh146, line 197: very thorough test data!

josh146 (commented):
A very solid PR @antalszava!

pennylane_orquestra/forest_device.py
antalszava: * `qe_component` is used to get the exact import for the framework
* `qe_module_name` and `qe_function_name` are used in specifying the `backend_specs` (see `create_backend_specs` method)

pennylane_orquestra/orquestra_device.py
antalszava: Yes, for now, they are assigned a `uuid`.

pennylane_orquestra/orquestra_device.py
antalszava: Good point, was also wondering about this: would the suggestion be to stick with the latest version of PennyLane or is there a feature in `>=0.13.0` to be used in the plugin? :thinking:

pennylane_orquestra/orquestra_device.py
antalszava: `"Hermitian"` is excluded from there (it requires `QubitUnitary` which is not supported when converting to OpenQASM), this is the original set in `QubitDevice`):
```python
    observables = {"PauliX", "PauliY", "PauliZ", "Hadamard", "Hermitian", "Identity"}
```

pennylane_orquestra/orquestra_device.py
josh146: Oh, isn't `batch_execute` only available in 0.13?

tests/test_orquestra_device.py
josh146: this is a very minor suggestion though, I'm happy for it to be left as is

pennylane_orquestra/orquestra_device.py
antalszava: True, but its subclasses (`QiskitDevice`, etc.) are instantiated. Getting a `TypeError: Can't instantiate abstract class B with abstract methods apply` for
```python
import pennylane as qml

class A(qml.QubitDevice):
    author = "Author"
    name = "Name"
    operations = None
    pennylane_requires = None
    short_name = None
    version = None

class B(A):
    pass

b = B()
```
Lmk if missing something though!

pennylane_orquestra/orquestra_device.py
antalszava: No, it is unsupported by the serializer, e.g.,:
```python
import pennylane as qml

dev = qml.device('orquestra.qulacs', wires=3)

@qml.qnode(dev)
def circuit():
    qml.PauliX(0).inv()
    return qml.expval(qml.PauliZ(0))

circuit()

~/xanadu/pennylane/pennylane/qnodes/base.py in decompose_queue(ops, device)
    127             raise qml.DeviceError(
    128                 "Gate {} not supported on device {}".format(op.name, device.short_name)
--> 129             ) from e
    130 
    131     return new_ops

DeviceError: Gate PauliX.inv not supported on device QASM serializer
```

pennylane_orquestra/orquestra_device.py
josh146: Oops. Got it!

pennylane_orquestra/orquestra_device.py
josh146: Oh, I see - none of the subclasses define `apply`. How come this is the case? I thought `apply` was required?

pennylane_orquestra/orquestra_device.py
antalszava: Sure, added more description!

pennylane_orquestra/orquestra_device.py
antalszava: Thanks for the heads-up!

pennylane_orquestra/orquestra_device.py
antalszava: That's a great point, forgot about it! Thanks for catching that :slightly_smiling_face:

pennylane_orquestra/orquestra_device.py
josh146: :(

pennylane_orquestra/orquestra_device.py
antalszava: Fair point, thank you! Updated

pennylane_orquestra/orquestra_device.py
antalszava: Not a bug per se, it's just that `decompose_hamiltonian` takes a matrix and not a PennyLane `Observable`. So when obtaining the matrix for the observable, wire information in lost by default. Perhaps we could think about changing `decompose_hamiltonian` or create a variant that can take an observable and preserve wiring internally.

pennylane_orquestra/utils.py
antalszava: Think that could be nice, though could the location break if the function is moved?

josh146: yep 😆

tests/test_orquestra_device.py
antalszava: That sounds like a good idea, could increase to 10000 shots for simulators and 8192 for IBMQ.

pennylane_orquestra/qulacs_device.py
antalszava: Oh, good point! Maybe good to keep now with increased shots? Perhaps for local purposes `shots=1024` makes sense and since we're using a simulator on a remote server, it's okay to have more shots.

pennylane_orquestra/forest_device.py
antalszava: Yes, fair point, will update the docstrings for each device. Thanks for catching that!

antalszava: Updated!

pennylane_orquestra/cli_actions.py
```diff
@@ -186,9 +186,11 @@ def loop_until_finished(workflow_id, timeout=300):
...
             current_status = workflow_details(workflow_id)
             raise TimeoutError(
-                f"The workflow results for workflow "
-                f"{workflow_id} were not obtained after {timeout/60} minutes. "
-                f"{current_status}"
+                "The workflow results for workflow "
+                f"{workflow_id} were not obtained after {timeout/60} minutes. \n"
+                "The timeout can be adjusted by specifying the 'timeout' "
+                "keyword argument.\n"
+                f"{''.join(current_status)}"
```
antalszava, line 193: Adjusted this message

antalszava:
Hi @josh146, thank you so much for the comments, they were very helpful! Think I've addressed them all, lmk.

pennylane_orquestra/ibmq_device.py
```diff
@@ -22,15 +22,36 @@
...
+            'q1', 'q2']``). Default 1 if not specified.
+        shots (int): number of circuit evaluations/random samples used to estimate
+            expectation values of observables
+        backend (str): the name of the Qiskit backend to use supported by
+            Orquestra, e.g., ``"ibmq_qasm_simulator"`` or the name of real hardware
+            devices
+
+    Keyword Args:
+        ibmqx_token=None (str): the authentication token needed to run a job on
+            IBMQ which is submitted to Orquestra as part of the workflow
```
josh146:
```suggestion
            IBMQ
```
(the orquestra part of the sentence seems a bit superfluous?)

josh146 (approved):
💯

pennylane_orquestra/orquestra_device.py
antalszava: The `OrquestraDevice.execute` and `OrquestraDevice.batch_execute` methods will be self-contained as they use circuit serialization on a `CircuitGraph` object as a whole, so it seemed that having an `apply` method definition might not make sense.

josh146: Got it, something to keep in mind for when we merge QubitDevice and Device

josh146: > The OrquestraDevice.execute and OrquestraDevice.batch_execute methods will be self-contained as they use circuit serialization on a CircuitGraph object as a whole, so it seemed that having an apply method definition might not make sense.

Perhaps worth adding the above as a comment to the `apply()` method, so that others developers understand why its defined but not used.

pennylane_orquestra/ibmq_device.py
```diff
@@ -0,0 +1,79 @@
...
+            set internally.
+    """
+
+    short_name = "orquestra.ibmq"
+
+    qe_component = "qe-qiskit"
+    qe_module_name = "qeqiskit.backend"
+    qe_function_name = "QiskitBackend"
+
+    def __init__(self, wires, shots=8192, backend="ibmq_qasm_simulator", **kwargs):
```
thisac, line 54: `shots=8192` :fire:

pennylane_orquestra/ibmq_device.py
```diff
@@ -0,0 +1,79 @@
...
+            raise ValueError(
+                "Please pass a valid IBMQX token to the device using the 'ibmqx_token' argument."
+            )
+
+        if kwargs.get("analytic", None):
+            # Raise a warning if the analytic attribute was set to True
+            warnings.warn(
+                f"The {self.short_name} device cannot be used in analytic "
+                "mode. Setting analytic to False, results are based on "
+                "sampling."
```
thisac: (minor) I would write this as two sentences instead.
```suggestion
                f"The {self.short_name} device cannot be used in analytic "
                "mode. Setting analytic to False. Results are based on "
                "sampling."
```

pennylane_orquestra/orquestra_device.py
```diff
@@ -0,0 +1,384 @@
...
+# limitations under the License.
+"""
+Base device class for PennyLane-Orquestra.
+"""
+# pylint: disable=protected-access, consider-using-enumerate
+
+import abc
+import json
+import re
+import uuid
```
thisac, line 22: These package names are really quite obscure. :laughing:

pennylane_orquestra/orquestra_device.py
```diff
@@ -0,0 +1,384 @@
...
+        backend=None (str): the Orquestra backend device to use for the
+            specific Orquestra backend, if applicable
+        batch_size=10 (int): the size of each circuit batch when using the
+            :meth:`~.batch_execute` method to send multiple workflows
+        keep_files=False (bool): Whether or not the workflow files
+            generated during the circuit execution should be kept or deleted.
+        resources=None (dict): An option for Orquestra, specifies the resources
+            provisioned for the clusters running each workflow step.
+        timeout=300 (int): The maximum time until a job will timeout after getting no
+            response from Orquestra (in seconds).
```
thisac: (minor) There seems to be a mix of capitalized-with-punctuation and not here. Maybe keep to one or the other; perhaps non-capitalized if only a single sentence? :thinking:

pennylane_orquestra/orquestra_device.py
```diff
@@ -0,0 +1,384 @@
...
+        Args:
+            circuit (~.CircuitGraph): circuit to serialize
+
+        Returns:
+            str: OpenQASM 2.0 representation of the circuit without any
+            measurement instructions
+        """
+        qasm_str = circuit.to_openqasm(rotations=not self.analytic)
+
+        qasm_without_measurements = re.sub("measure.*?;\n?\s*", "", qasm_str)
```
thisac, line 236: :smile:

pennylane_orquestra/orquestra_device.py
```diff
@@ -0,0 +1,384 @@
...
+        Args:
+            wires (Wires): the wires the observable of the quantum function
+                acts on
+
+        Returns:
+            str: the ``openfermion.IsingOperator`` string representation
+        """
+        op_wires_but_last = [f"Z{w} " for w in wires[:-1]]
+
+        # No space after the last wire
```
thisac, line 315: :+1:

thisac (approved):
Looks great Antal. Only some very minor comment; mainly docs related. :100:

pennylane_orquestra/orquestra_device.py
antalszava: Fair point, updated! Thanks

antalszava:
@josh146 @thisac Thanks for the comments! Merging (docs not yet available in this PR).

PR #6, opened 2020-12-16
Add the execute and batch_execute methods

antalszava:
**Context**
`OrquestraDevice` and its subclasses are capable of serializing backend specifications, circuits and operators. There are also auxiliary functions available that make it possible to:
* fit serialized data into an Orquestra workflow (`gen_workflow.py`)
* place CLI calls to Orquestra (`cli_actions.py`).

To put pieces together, executing circuits and computing expectation values of operators remain.

**Changes**
Adds two device methods:
* `OrquestraDevice.execute`: supports computing expectation values for (an) operator(s) and returning the result(s) to a QNode
* `OrquestraDevice.batch_execute`: supports computing expectation values for batches of circuits (e.g., when computing jacobians and using a `JacobianTape` with the new tape mode feature of PennyLane)
* End-to-end integration tests that test the connectivity of a local machine with Orquestra workflow submission (just like a user would use the available devices). These tests are not run by the CI.


steps/test_expval.py
```diff
@@ -160,7 +160,6 @@ def test_run_circuit_and_get_expval_rot_cnot(self, backend_specs, monkeypatch):
         res = np.array(lst[0])
         assert np.allclose(res, np.array([np.cos(theta), np.cos(theta) * np.cos(phi)]), atol=tol)
 
-    @pytest.mark.xfail
```
antalszava, line 163: Can be removed as devices can execute circuits now

pennylane_orquestra/orquestra_device.py
```diff
@@ -233,7 +238,7 @@ def serialize_circuit(self, circuit):
         """
         qasm_str = circuit.to_openqasm(rotations=not self.analytic)
 
-        qasm_without_measurements = re.sub("measure.*?;\n?\s*", "", qasm_str)
+        qasm_without_measurements = re.sub(r"measure.*?;\n?\s*", "", qasm_str)
```
antalszava, line 241: modification due to a warning

.github/workflows/step_tests.yml
```diff
@@ -39,6 +39,11 @@ jobs:
           pip install -r steps/steps_requirements.txt
           pip install wheel pytest pytest-cov pytest-mock flaky --upgrade
 
+      - name: Install PennyLane-Orquestra
+        run: |
+          python setup.py bdist_wheel
+          pip install dist/PennyLane_Orquestra*.whl
+
```
antalszava, line 46: Can use the plugin now here ([see this comment](https://github.com/PennyLaneAI/pennylane-orquestra/pull/6/files#r543933649))

pennylane_orquestra/gen_workflow.py
```diff
@@ -63,13 +63,13 @@ def step_dictionary(name_suffix):
             "runtime": {
                 "language": "python3",
                 "imports": [
-                    "pl_orquestra",
+                    "pennylane_orquestra",
```
antalszava, line 66: With #2 merged, the steps can be redirected to this repo :champagne:

tests/conftest.py
```diff
@@ -85,9 +85,9 @@ def custom_wires(request):
 first_config = {
     "runtime": {
         "language": "python3",
-        "imports": ["pl_orquestra", "z-quantum-core", "qe-openfermion", "qe-forest"],
+        "imports": ["pennylane_orquestra", "z-quantum-core", "qe-openfermion", "qe-forest"],
```
antalszava, line 96: Also due to merging #2 and using steps from this repo

tests/test_orquestra_device.py
```diff
@@ -189,13 +312,20 @@ def circuit():
...
+        "0.4999999999999999 [X2 X1] + 0.4999999999999999 [X2 Z1] + 0.4999999999999999 [Z2 X1] + 0.4999999999999999 [Z2 Z1]",
+    ),
+    (
+        qml.Hermitian((qml.PauliY(1) @ qml.PauliX(0) @ qml.PauliZ(2)).matrix, wires=[1, 0, 2]),
+        "1.0 [Y1 X0 Z2]",
+    ),
+    (qml.PauliY(2) @ qml.Hermitian((qml.PauliX(1)).matrix, wires=[1]), "1.0 [Y2 X1]"),
 ]
 
+
```
antalszava, line 328: Ran `black`

.github/workflows/tests.yml
```diff
@@ -46,7 +46,7 @@ jobs:
           pip install dist/PennyLane_Orquestra*.whl
 
       - name: Run tests
-        run: python -m pytest tests $COVERAGE_FLAGS
+        run: python -m pytest tests -k 'not e2e' $COVERAGE_FLAGS
```
josh146, line 49: nice!

pennylane_orquestra/forest_device.py
```diff
@@ -15,6 +15,7 @@
 The Forest device class for PennyLane-Orquestra.
 """
 from pennylane_orquestra.orquestra_device import OrquestraDevice
+import warnings
```
josh146, line 18: How come this change was made?

pennylane_orquestra/orquestra_device.py
```diff
@@ -382,3 +387,254 @@ def qubit_operator_string(self, observable):
...
         wire_map = {v: idx for idx, v in enumerate(self.wires)}
         return _terms_to_qubit_operator_string(coeffs, obs_list, wires=wire_map)
+
+    def execute(self, circuit, **kwargs):
+        # Input checks
+        not_all_expval = any(obs.return_type is not Expectation for obs in circuit.observables)
+        if not_all_expval:
+            raise NotImplementedError(
+                f"The {self.short_name} device only supports returning expectation values."
+            )
```
josh146, line 397: Good idea to check this. The other alternative (not defining the `var` and `sample` methods) would result in an error message only occurring _after_ the job returns from the server (if it gets that far).

pennylane_orquestra/orquestra_device.py
```diff
@@ -382,3 +387,254 @@ def qubit_operator_string(self, observable):
...
+
+        res = self._asarray(results)
+        return res
+
+    def single_step_results(self, workflow_id):
+        """Extracts the results of a single step obtained for a workflow.
+
+        This method assumes that the workflow had a single step and that the
+        structure of the result corresponds to results sent by Orquestra API
+        v1.0.0.
```
josh146, line 458: Is there a way to verify this? Perhaps the API version can be checked, and a warning emitted if it is higher than 1.0.0?

pennylane_orquestra/orquestra_device.py
```diff
@@ -382,3 +387,254 @@ def qubit_operator_string(self, observable):
...
+        data = loop_until_finished(workflow_id, timeout=self._timeout)
+        try:
+            step_result = [v for k, v in data.items()][0]
+            results = step_result["expval"]["list"]
+        except (IndexError, KeyError, TypeError, AttributeError) as e:
+            current_status = workflow_details(workflow_id)
+            raise ValueError(
+                f"Unexpected result format for workflow {workflow_id}.\n "
+                "{''.join(current_status)}"
+            )
```
josh146:
```suggestion
            ) from e
```

pennylane_orquestra/orquestra_device.py
```diff
@@ -382,3 +387,254 @@ def qubit_operator_string(self, observable):
...
+        using workflows into batch results.
+
+        Computations involving the identity observable are given by theoretical
+        values rather than as part of a workflow. Therefore, such values need
+        to be inserted into the results later.
+
+        Args:
+            results (list): workflow results of the batched execution
+            empty_obs_list (list): list of indices where every observable was
+                identity
```
josh146:
```suggestion
            empty_obs_list (list): list of indices where every observable is the identity
```

pennylane_orquestra/orquestra_device.py
```diff
@@ -382,3 +387,254 @@ def qubit_operator_string(self, observable):
...
+            identity_indices (dict): maps the index of a sublist to the
+                the list of indices where the observable was an identity
```
josh146:
```suggestion
                the list of indices where the observable is an identity
```

pennylane_orquestra/orquestra_device.py
```diff
@@ -382,3 +387,254 @@ def qubit_operator_string(self, observable):
...
+
+        Returns:
+            list: list of results
+        """
+        # Insert the lists needed for only identity results
+        for idx in empty_obs_list:
+            results.insert(idx, [])
```
josh146, line 498: Since we are mutating the `results` list as part of the loop, will this cause an inconsistency (e.g, will the correct `idx` start to 'drift'?)

pennylane_orquestra/orquestra_device.py
```diff
@@ -382,3 +387,254 @@ def qubit_operator_string(self, observable):
...
+
+        # Insert further identity results
+        for list_idx in identity_indices.keys():
+            for iden_idx in identity_indices[list_idx]:
+                results[list_idx].insert(iden_idx, 1)
```
josh146, line 503: Same question here!

pennylane_orquestra/orquestra_device.py
```diff
@@ -382,3 +387,254 @@ def qubit_operator_string(self, observable):
...
+        # circuits per workflow
+        while idx < len(circuits):
+            end_idx = idx + self._batch_size
+            batch = circuits[idx:end_idx]
+            file_id = f"{file_prefix}-{str(idx)}"
+
+            res = self._batch_execute(batch, file_id, **kwargs)
+
+            results.extend(res)
+            idx += self._batch_size
```
josh146: 🍾

pennylane_orquestra/orquestra_device.py
```diff
@@ -382,3 +387,254 @@ def qubit_operator_string(self, observable):
...
+
+        return results
+
+    def _batch_execute(self, circuits, file_id, **kwargs):
```
josh146: maybe it would be good to give this a more descriptive name, so that it is more easily distinguished from `batch_execute`? Maybe `multi_step_execute`?

pennylane_orquestra/orquestra_device.py
```diff
@@ -382,3 +387,254 @@ def qubit_operator_string(self, observable):
...
+                raise NotImplementedError(
+                    f"The {self.short_name} device only supports returning expectation values."
+                )
+
+            self.check_validity(circuit.operations, circuit.observables)
+
+        # 1. Create qasm strings from the circuits
+        # Extract the CircuitGraph object from QuantumTape
+        circuits = [circ.graph for circ in circuits]
+        qasm_circuits = [self.serialize_circuit(circuit) for circuit in circuits]
```
josh146, line 550: Here, I guess the `circuits` is _always_ a tape

pennylane_orquestra/orquestra_device.py
```diff
@@ -382,3 +387,254 @@ def qubit_operator_string(self, observable):
...
+            data = dict(sorted(data.items(), key=get_step_name))
+
+            # Obtain the results for each step
+            result_dicts = [v for k, v in data.items()]
+            results = [dct["expval"]["list"] for dct in result_dicts]
+        except (IndexError, KeyError, TypeError, AttributeError) as e:
+            current_status = workflow_details(workflow_id)
+            raise ValueError(
+                f"Unexpected result format for workflow {workflow_id}.\n "
+                "{''.join(current_status)}"
```
josh146:
```suggestion
                f"{''.join(current_status)}"
```

pennylane_orquestra/orquestra_device.py
```diff
@@ -382,3 +387,254 @@ def qubit_operator_string(self, observable):
...
+        """
+        data = loop_until_finished(workflow_id, timeout=self._timeout)
+        try:
+            step_result = [v for k, v in data.items()][0]
+            results = step_result["expval"]["list"]
+        except (IndexError, KeyError, TypeError, AttributeError) as e:
+            current_status = workflow_details(workflow_id)
+            raise ValueError(
+                f"Unexpected result format for workflow {workflow_id}.\n "
+                "{''.join(current_status)}"
```
josh146:
```suggestion
                f"{''.join(current_status)}"
```

pennylane_orquestra/qiskit_device.py
```diff
@@ -52,9 +52,20 @@ def __init__(self, wires, shots=10000, backend="qasm_simulator", **kwargs):
                 # Raise a warning if the analytic attribute was set to True
                 warnings.warn(
                     f"The {self.short_name} device cannot be used in analytic "
-                    "mode. Setting analytic to False. Results are based on "
-                    "sampling."
+                    "mode with the {backend} backend. Setting analytic to False. "
```
josh146:
```suggestion
                    f"mode with the {backend} backend. Setting analytic to False. "
```

pennylane_orquestra/qiskit_device.py
```diff
@@ -52,9 +52,20 @@ def __init__(self, wires, shots=10000, backend="qasm_simulator", **kwargs):
...
+
+        # TODO: Remove when the Orquestra supports qiskit>0.18.3 with its Qiskit component
+        if backend == "statevector_simulator":
+            if not kwargs.get("analytic", True):
+                # Raise a warning if the analytic attribute was set to False
+                warnings.warn(
+                    f"The {self.short_name} device with the {backend} backend "
+                    "always runs with shots=1 due to a malfunction in the "
+                    "version of Qiskit used by Orquestra."
+                )
```
josh146, line 69: 😨

tests/conftest.py
```diff
@@ -195,6 +195,45 @@ def custom_wires(request):
 step_name1 = "run-circuit-and-get-expval-1"
 step_name2 = "run-circuit-and-get-expval-2"
 
+@pytest.fixture()
+def test_batch_result():
+    """Example batch results used in tests.
+
+    The step names are not in order ("stepName" entry in each nested
+    dictionary), so that the order of results differs from the way they were
+    assumed to be submitted and ordering can be tested too.
```
josh146, line 212: Nice!

tests/conftest.py
```diff
@@ -211,3 +250,13 @@ def test_result():
     }
 
     return test_res
+
+@pytest.fixture
+def token():
+    """Get the IBMQX token from an environment variable."""
+    t = os.getenv("IBMQX_TOKEN_TEST", None)
```
josh146: @antalszava I named this `IBMQX_TOKEN` in the repository secret!

josh146: or is the plan not to use this for CI anyway?

tests/test_e2e_integration.py
```diff
@@ -0,0 +1,241 @@
...
+qiskit_analytic_specs = '{"module_name": "qeqiskit.simulator", "function_name": "QiskitSimulator", "device_name": "qasm_simulator"}'
+qiskit_sampler_specs = '{"module_name": "qeqiskit.simulator", "function_name": "QiskitSimulator", "device_name": "qasm_simulator", "n_samples": 1000}'
+
+analytic_tol = 10e-10
+
+# The tolerance for sampling is expected to be higher
+tol = 10e-2
+
+# Expected to fail if qe is unavailable (for example for the CI)
+@pytest.mark.xfail(raises=FileNotFoundError)
```
josh146: Is this because qe is not installed on the CI?

tests/test_e2e_integration.py
```diff
@@ -0,0 +1,241 @@
...
+        default_qnode = qml.QNode(func, dev2)
+
+        dfunc1 = qml.grad(orquestra_qnode)
+        dfunc2 = qml.grad(default_qnode)
+
+        res_orquestra = dfunc1(weights)
+        res_default_qubit = dfunc2(weights)
+
+        assert np.allclose(res_orquestra, res_default_qubit)
+        qml.disable_tape()
```
josh146: @antalszava better to make a fixture for this, so that tape mode is always disabled even if the test fails middway.

```python
@pytest.fixture
def tape_mode():
    qml.enable_tape()
    yield
    qml.disable_tape()
```

tests/test_orquestra_device.py
```diff
@@ -84,6 +101,112 @@ def test_empty_apply(self):
...
+            @qml.qnode(dev)
+            def circuit():
+                qml.PauliX(0)
+                return qml.expval(qml.PauliZ(0))
+
+            assert circuit() == 123456789
+
+            # Check that the resorces were passed correctly
+            assert len(recorder) == 1
+            assert recorder[0] == resources
```
josh146, line 207: Really nice testing

tests/test_orquestra_device.py
```diff
@@ -291,3 +421,418 @@ def circuit():
...
+                pennylane_orquestra.orquestra_device,
+                "loop_until_finished",
+                lambda *args, **kwargs: test_batch_result,  # The exact results are not considered in the test
+            )
+
+            dev.execute(tape1)
+
+        expected = 'OPENQASM 2.0;\ninclude "qelib1.inc";\nqreg q[1];\ncreg c[1];\nh q[0];\nry(-0.7853981633974483) q[0];\n'
+        assert circuit_history[0] == expected
+        qml.disable_tape()
```
josh146: Same here re:fixture

tests/test_orquestra_device.py
```diff
@@ -291,3 +421,418 @@ def circuit():
...
+                pennylane_orquestra.orquestra_device,
+                "loop_until_finished",
+                lambda *args, **kwargs: test_batch_result,  # The exact results are not considered in the test
+            )
+
+            dev.execute(tape1)
+
+        expected = 'OPENQASM 2.0;\ninclude "qelib1.inc";\nqreg q[1];\ncreg c[1];\nh q[0];\n'
+        assert circuit_history[0] == expected
+        qml.disable_tape()
```
josh146: same here

tests/test_orquestra_device.py
```diff
@@ -291,3 +421,418 @@ def circuit():
...
+
+            for r, e in zip(res, expected):
+                assert np.allclose(r, e)
+
+        qml.disable_tape()
+
+    @pytest.mark.parametrize("mock_res", [[], "res", {"a": 3}, {"a": []}])
+    def test_wrong_result_format_multiple_steps(self, mock_res, monkeypatch):
+        """Test that the resource details defined when the device was created
+        are passed to generate the workflow."""
```
josh146: Does this docstring match the test?

josh146 (approved):
This is very exciting @antalszava! Have left mostly questions, but also a few suggestions.

pennylane_orquestra/forest_device.py
antalszava: It was a missed import, added a test case too using `pytest.mark.parametrize` to cover this:
```python
non_analytic_backends = [
    ("orquestra.qiskit", "qasm_simulator"),
    ("orquestra.forest", "3q-noisy-qvm"),
]

@pytest.mark.parametrize("dev,backend", non_analytic_backends)
def test_analytic_warning(self, dev, backend):
```
Edit: adjusted the test name and docstring.

pennylane_orquestra/orquestra_device.py
antalszava: Good idea, changed! Thanks

pennylane_orquestra/orquestra_device.py
antalszava: Yes, as per the original `batch_execute` method (which takes `circuits (list[.tapes.QuantumTape])`). Added a comment to clarify.

tests/conftest.py
antalszava: Oh was wondering about it, thanks! Updated the names to use `IBMQX_TOKEN`.

tests/test_e2e_integration.py
antalszava: Yes, although on second thought it might be overkill as the CI skips it explicitly. Will rather remove these, so that if the `e2e` tests are run explicitly, they are not unexpectedly passing (`X`).

tests/test_e2e_integration.py
antalszava: Oh nice, that's a good one! Actually just realized that these are all unit tests which construct tapes explicitly, so tape mode doesn't seem to be needed. However, changed to using this fixture in the relevant `e2e` test case.

tests/test_orquestra_device.py
antalszava: Ooops, no, good catch! Thank you, updated

pennylane_orquestra/orquestra_device.py
antalszava: Hmm, good question! `qe --version` seems to output only the `Client` and `Server` versions :thinking:

.github/workflows/tests.yml
```diff
@@ -46,7 +46,7 @@ jobs:
...
 
       - name: Adjust coverage file for Codecov
         run: bash <(sed -i 's/filename=\"/filename=\"pennylane_orquestra\//g' coverage.xml)
```
thisac: Again, just re-iterating what @josh146 said earlier, is this necessary here?

pennylane_orquestra/gen_workflow.py
```diff
@@ -115,11 +115,11 @@ def gen_expval_workflow(component, backend_specs, circuits, operators, **kwargs)
...
         "imports": [
             {
-                "name": "pl_orquestra",
+                "name": "pennylane_orquestra",
                 "type": "git",
                 "parameters": {
-                    "repository": "git@github.com:antalszava/pl_orquestra.git",
-                    "branch": "master",
+                    "repository": "git@github.com:PennyLaneAI/pennylane-orquestra.git",
+                    "branch": "main",
```
thisac, line 122: Is "master" now called "main" or are both accepted still?

pennylane_orquestra/orquestra_device.py
```diff
@@ -382,3 +387,254 @@ def qubit_operator_string(self, observable):
...
+        except AttributeError:
+            # QuantumTape case: need to extract the CircuitGraph
+            qasm_circuit = self.serialize_circuit(circuit.graph)
+
+        # 2. Create the qubit operators
+        ops, identity_indices = self.process_observables(circuit.observables)
+
+        if not ops:
+            # All the observables were identity, no workflow submission needed
+            return self._asarray([1] * len(identity_indices))
```
thisac, line 413: :tada:

pennylane_orquestra/orquestra_device.py
antalszava: Hi @josh146, that's a great question! Actually was also wondering about this when adding it. One thing to note is that `empty_obs_list` is special in two ways: a) it is sorted in a strictly ascending order b) it stores indices with regards to the complete list (not a partial version).

In that sense, `idx` can be:
a) `idx==0` (inserting to the head of the list, but only once)
b) in-between existing elements (e.g., if `idx<len(results)`)
c) last (e.g., if `idx==len(results)`)

*Why is this fine for each case?*
a) and c) are fine, as we are always appending to either the head or the tail of the list using indices of the complete list.

b) Since `empty_obs_list` is sorted, we can be sure that for the next `idx` the sublist `results[:idx]` is already complete, as no insertion will happen to an `i<idx` index. So basically after insertion, `results[:idx+1]` will not be mutated.

The second half of the list, `results[idx+1:]` is still incomplete this time and elements in `results[idx+1:]` are not necessarily at their correct indices yet and can be "pushed back".

*Example*

Creating the `[[], [], [], 1, [], 2]` list
```python
In [8]: results = [1,2]  
In [10]: empty_obs_list = [0, 1,2, 4]
In [11]: for idx in empty_obs_list: 
    ...:       results.insert(idx, []) 
    ...:   
In [12]: results                                                                                                                                  
Out[12]: [[], [], [], 1, [], 2]
```

josh146: Thanks Antal! Good explanation

pennylane_orquestra/orquestra_device.py
```diff
@@ -382,3 +387,254 @@ def qubit_operator_string(self, observable):
...
+
+            if not any(ops):
+                # All the batches only had identity observables, no workflow submission needed
+                return [self._asarray([1] * len(circuit.observables)) for circuit in circuits]
+
+            # Remove the empty lists so that those are not submitted
+            ops = [o for o in ops if o]
+
+        # Multiple steps: need to create json strings as elements of the list
+        ops = [json.dumps(o) for o in ops]
```
thisac, line 579: Very nice comments everywhere. Easy to follow! :tada:

pennylane_orquestra/orquestra_device.py
```diff
@@ -382,3 +387,254 @@ def qubit_operator_string(self, observable):
...
+
+        filename = f"expval-{file_id}.yaml"
+        filepath = write_workflow_file(filename, workflow)
+
+        # 5. Submit the workflow
+        workflow_id = qe_submit(filepath, keep_file=self._keep_files)
+        self._latest_id = workflow_id
+
+        if self._keep_files:
+            self._filenames.append(filename)
```
thisac, line 599: Are all files that are not in `self._filenames` deleted somewhere? I might've missed/forgotten where that was done.

pennylane_orquestra/orquestra_device.py
```diff
@@ -382,3 +387,254 @@ def qubit_operator_string(self, observable):
...
+            workflow_id (str): the ID of the workflow to extract results for
+
+        Returns:
+            results (list): a list of workflow results for each step
+        """
+        data = loop_until_finished(workflow_id, timeout=self._timeout)
+        try:
+            # Sort results by step name
+            get_step_name = lambda entry: entry[1]["stepName"]
+            data = dict(sorted(data.items(), key=get_step_name))
```
thisac, line 631: Why do you need to sort by step name?

tests/test_e2e_integration.py
```diff
@@ -0,0 +1,232 @@
...
+    @pytest.mark.parametrize("resources", [None, resources_default])
+    def test_can_submit_and_query_workflow_details(self, resources, tmpdir):
+        """Test that filling in the workflow template for getting expectation
+        values can be submitted to Orquestra and workflow details can be queried."""
+        # Skip if not logged in to Orquestra
+        try_resp = qe_list_workflow()
+        need_login_msg = "token has expired, please log in again\n"
+
+        if need_login_msg in try_resp:
+            pytest.skip("Has not logged in to the Orquestra platform.")
```
thisac, line 61: Nice!

tests/test_e2e_integration.py
```diff
@@ -0,0 +1,232 @@
...
+        workflow = gw.gen_expval_workflow(
+            backend_component, backend_specs_default, circuits, op, resources=resources
+        )
+        file_name = tmpdir.join("test_workflow.yaml")
+
+        with open(file_name, "w") as file:
+            d = yaml.dump(workflow, file)
+
+        # Submit a workflow
+        workflow_id = qe_submit(file_name)
```
thisac: Just to clarify: this creates a workflow, stores it in a new file called "test_workflow.yaml" and submits that file, right? Probably not an issue, but should we remove that file after the test is finished?

tests/test_e2e_integration.py
```diff
@@ -0,0 +1,232 @@
...
+        circuits = [qasm_circuit_default]
+
+        # Fill in workflow template
+        workflow = gw.gen_expval_workflow(
+            backend_component, backend_specs_default, circuits, op, resources=resources
+        )
+        file_name = tmpdir.join("test_workflow.yaml")
+
+        with open(file_name, "w") as file:
+            d = yaml.dump(workflow, file)
```
thisac: (minor)
```suggestion
            yaml.dump(workflow, file)
```

tests/test_e2e_integration.py
```diff
@@ -0,0 +1,232 @@
...
+        operator = []
+
+        # Fill in workflow template
+        workflow = gw.gen_expval_workflow(
+            backend_component, backend_specs_default, circuits, operator
+        )
+        file_name = tmpdir.join("test_workflow.yaml")
+
+        with open(file_name, "w") as file:
+            d = yaml.dump(workflow, file)
```
thisac: (minor)
```suggestion
            yaml.dump(workflow, file)
```

tests/test_orquestra_device.py
```diff
@@ -50,24 +54,36 @@ def circuit():
...
             match="cannot be used in analytic",
         ):
-            dev = qml.device("orquestra.qiskit", backend="qasm_simulator", wires=2, analytic=True)
+            dev = qml.device(dev, backend=backend, wires=2, analytic=True)
+
+        assert not dev.analytic
+
+    def test_statevector_simulator_single_shot_warning(self):
+        """Test that a warning is raised when using the QeQiskitDevice with the
+        statevector_simulator backend in sampling mode that the shots=1."""
```
thisac: This needs a rewrite (I think you mixed two sentences into one)!

tests/test_orquestra_device.py
```diff
@@ -84,6 +100,112 @@ def test_empty_apply(self):
...
+            m.setattr(uuid, "uuid4", lambda *args: test_uuid)
+
+            @qml.qnode(dev)
+            def circuit():
+                qml.PauliX(0)
+                return qml.expval(qml.PauliZ(0))
+
+            assert circuit() == 123456789
+            file_kept = os.path.exists(tmpdir.join(f"expval-{test_uuid}.yaml"))
+            assert file_kept if keep else not file_kept
```
thisac, line 136: :fire:

tests/test_orquestra_device.py
```diff
@@ -84,6 +100,112 @@ def test_empty_apply(self):
...
+            assert dev.filenames == ([f"expval-{test_uuid}.yaml"] if keep else [])
+            assert dev.latest_id == "SomeWorkflowID"
```
thisac: Where does the "SomeWorkflowID" name come from? :thinking:

tests/test_orquestra_device.py
```diff
@@ -84,6 +100,112 @@ def test_empty_apply(self):
...
+
+    @pytest.mark.parametrize("timeout", [1, 2.5])
+    def test_timeout(self, timeout, tmpdir, monkeypatch):
+        """Test the option for keeping/deleting the workflow file."""
```
thisac: Same docstring as above. Should probably be updated!

tests/test_orquestra_device.py
```diff
@@ -84,6 +100,112 @@ def test_empty_apply(self):
...
+            @qml.qnode(dev)
+            def circuit():
+                qml.PauliX(0)
+                return qml.expval(qml.PauliZ(0))
+
+            start = time.time()
+            with pytest.raises(TimeoutError, match="The workflow results for workflow"):
+                circuit()
+            end = time.time()
+            assert end - start >= timeout
```
thisac, line 170: Is this test simply reliant on that the calling of `circuit()` will take longer than `timeout` (in this case 1 and 2.5 seconds)? What if it's run on a very fast computer? Might it fail then? :thinking:

tests/test_orquestra_device.py
```diff
@@ -291,3 +420,393 @@ def circuit():
...
             ):
                 circuit()
+
+
+class TestExecute:
+    """Tests for the execute method of the base OrquestraDevice class."""
+
+    def test_serialize_circuit_rotations_tape(self, monkeypatch, tmpdir, test_batch_result):
+        """Test that a circuit that is serialized correctly with rotations for
+        a remote hardware backend in tape mode"""
```
thisac:
```suggestion
        """Test that a circuit is serialized correctly with rotations for
        a remote hardware backend in tape mode."""
```

tests/test_orquestra_device.py
```diff
@@ -291,3 +420,393 @@ def circuit():
...
+            )
+
+            dev.execute(tape1)
+
+        expected = 'OPENQASM 2.0;\ninclude "qelib1.inc";\nqreg q[1];\ncreg c[1];\nh q[0];\nry(-0.7853981633974483) q[0];\n'
+        assert circuit_history[0] == expected
+
+    def test_serialize_circuit_no_rotations_tape(self, monkeypatch, tmpdir, test_batch_result):
+        """Test that a circuit that is serialized correctly without rotations for
+        a simulator backend in tape mode"""
```
thisac:
```suggestion
        """Test that a circuit is serialized correctly without rotations for
        a simulator backend in tape mode."""
```

tests/test_orquestra_device.py
```diff
@@ -291,3 +420,393 @@ def circuit():
...
+        def circuit():
+            qml.PauliX(0)
+            return qml.expval(qml.Identity(0)), qml.expval(qml.Identity(1))
+
+        assert np.allclose(circuit(), np.ones(2))
+
+    @pytest.mark.parametrize("dev", ["orquestra.forest", "orquestra.qiskit", "orquestra.qulacs"])
+    def test_identity_mixed(self, dev, monkeypatch, tmpdir, test_result):
+        """Test computing that computing the expectation value of the identity
+        and PauliZ returns an array of results."""
```
thisac:
```suggestion
        """Test that computing the expectation values of the identity
        and PauliZ returns an array of results."""
```

tests/test_orquestra_device.py
```diff
@@ -291,3 +420,393 @@ def circuit():
...
+                qml.PauliX(0)
+                return qml.expval(qml.Identity(0)), qml.expval(qml.PauliZ(1))
+
+            res = circuit()
+            assert np.allclose(res, np.array([1, test_batch_res0]))
+
+    @pytest.mark.parametrize("mock_res", [[], "res", {"a": 3}, {"a": []}])
+    def test_wrong_result_format_single_step(self, mock_res, monkeypatch):
+        """Test that if a result with unexpected format of a single-step
+        workflow is obtained then an error is raised."""
```
thisac:
```suggestion
        """Test that an error is raised if a result with unexpected format of a single-step
        workflow is obtained."""
```

tests/test_orquestra_device.py
```diff
@@ -291,3 +420,393 @@ def circuit():
...
+
+            with pytest.raises(ValueError, match="Unexpected result format"):
+                dev.single_step_results("SomeID")
+
+class TestBatchExecute:
+    """Test the integration of the device with PennyLane."""
+
+    def test_error_if_not_expval_batched(self):
+        """Test that an error is raised if not an expectation value is
+        computed during batched execution"""
```
thisac:
```suggestion
        """Test that an error is raised if an expectation value isn't
        computed during batched execution."""
```

tests/test_orquestra_device.py
```diff
@@ -291,3 +420,393 @@ def circuit():
...
+            )
+            m.setattr(
+                pennylane_orquestra.orquestra_device, "workflow_details", lambda *args: "SomeStatus"
+            )
+
+            with pytest.raises(ValueError, match="Unexpected result format"):
+                dev.single_step_results("SomeID")
+
+class TestBatchExecute:
+    """Test the integration of the device with PennyLane."""
```
thisac: Should this rather be something like "Test batch execution ..."?

tests/test_orquestra_device.py
```diff
@@ -291,3 +420,393 @@ def circuit():
...
+
+        res = dev.batch_execute([tape1])
+        assert len(res) == 1
+        assert np.allclose(res[0], np.array([1]))
+
+    @pytest.mark.parametrize("dev", ["orquestra.forest", "orquestra.qiskit", "orquestra.qulacs"])
+    def test_identity_mixed(self, dev, monkeypatch, tmpdir, test_result):
+        """Test computing that computing the expectation value of a tape with
+        observables of identity and PauliZ and a tape where only the identity
+        of observable returns the correct list of results."""
```
thisac: This line is a bit weird. Should probably be updated!

tests/test_orquestra_device.py
```diff
@@ -291,3 +420,393 @@ def circuit():
...
+                qml.expval(qml.Identity(wires=[0]))
+                qml.expval(qml.PauliZ(wires=[1]))
+
+            with qml.tape.QuantumTape() as tape2:
+                qml.expval(qml.Identity(wires=[0]))
+
+            res = dev.batch_execute([tape1, tape2])
+
+            assert np.allclose(res[0], np.array([1, test_batch_res0]))
+            assert np.allclose(res[1], np.array([1]))
```
thisac, line 625: :fire:

tests/test_orquestra_device.py
```diff
@@ -291,3 +420,393 @@ def circuit():
...
+
+            res = dev.batch_execute(circuits)
+
+            # Correct order of results is expected
+            assert np.allclose(res[0], test_batch_res0)
+            assert np.allclose(res[1], test_batch_res1)
+            assert np.allclose(res[2], test_batch_res2)
+            file_kept = os.path.exists(tmpdir.join(f"expval-{test_uuid}-0.yaml"))
+
+            assert file_kept if keep else not file_kept
```
thisac, line 687: I see you snuck in an extra test for keeping files here. :wink:

thisac (approved):
Very nice code and testing @antalszava. Just had a few questions and some text-comments (a few docstrings in the tests need to be updated, etc.) :medal_sports:

pennylane_orquestra/gen_workflow.py
josh146: Any new GitHub repo created, the default branch name will be `main`. Existing repos will retain a default `master` branch, unless we want to change it manually.

.github/workflows/tests.yml
antalszava: Oh, thanks so much for catching that! Still remained there somehow (maybe in between branches). Removed.

pennylane_orquestra/orquestra_device.py
antalszava: That's a great question! It was added in #3, an addition that might more so require access to `qe`. It is taken care in [`qe_submit` by passing the boolean `keep_file`](https://github.com/PennyLaneAI/pennylane-orquestra/blob/a654fb207e1448f077f57b74dd44eac84bcb6e16/pennylane_orquestra/cli_actions.py#L83).

pennylane_orquestra/orquestra_device.py
antalszava: Detailed in the docstring:
>         Due to parallel execution, results might have been written in any
>         order, so results are sorted by the step name.

Orquestra doesn't necessarily execute parallel steps in the order they were defined in a workflow file (adding this bit too).

tests/test_e2e_integration.py
antalszava: Good point! `keep_file=False` is the default for `qe_submit`, so the file is removed.

tests/test_orquestra_device.py
antalszava: Good catch, thanks! Updated.

tests/test_orquestra_device.py
antalszava: Was in `conftest`, though pretty hidden. Created `test_workflow_id_name` and now that's imported here.

tests/test_orquestra_device.py
antalszava: Yep!

tests/test_orquestra_device.py
antalszava: The test is created so that [this while loop](https://github.com/PennyLaneAI/pennylane-orquestra/blob/a654fb207e1448f077f57b74dd44eac84bcb6e16/pennylane_orquestra/cli_actions.py#L182) is an infinite loop (due to an `IndexError` being raised everytime for `"Test res"` when a location is attempted to be parsed). Therefore raising the `TimeoutError` is interrupting the infinite loop.

PR #11, opened 2020-12-21
Correct plugin links in readme

antalszava:


josh146 (approved):
:+1: good catch. Does the incorrect link occur anywhere else? Might be worth searching the repo

antalszava:
Fortunately, these were the only two places. Had a go with search & replace on the whole repo.

PR #14, opened 2020-12-21
Updates IBMQ no token test

antalszava:
1. Updates the IBMQ device no token test to explicitly exclude any environment variables during the test case.
2. Expands the error message raised when no token was found.

### DIFF
--- a/tests/test_orquestra_device.py
+++ b/tests/test_orquestra_device.py
@@ -89,9 +89,10 @@ class TestBaseDevice:
 
         assert not dev.analytic
 
-    def test_ibmq_no_token_error(self):
+    def test_ibmq_no_token_error(self, monkeypatch):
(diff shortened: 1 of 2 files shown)


tests/test_orquestra_device.py
```diff
@@ -89,9 +89,10 @@ def test_ibmq_analytic_warning(self):
 
         assert not dev.analytic
 
-    def test_ibmq_no_token_error(self):
+    def test_ibmq_no_token_error(self, monkeypatch):
         """Test that an error is raised when using the IBMQDevice without any
         tokens specified."""
+        monkeypatch.delenv("IBMQX_TOKEN", raising=False)
```
josh146, line 95: Good catch, this is very subtle!

josh146 (approved):

PR #17, opened 2021-01-28
Tests in tape mode

antalszava:
Adjusts some tests to be compatible with tape mode.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
-git+https://github.com/PennyLaneAI/pennylane.git@tape_hash
+git+https://github.com/PennyLaneAI/pennylane.git
 pyyaml
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
-pennylane>=0.13.0
+git+https://github.com/PennyLaneAI/pennylane.git@tape_hash
 pyyaml
--- a/steps/expval.py
+++ b/steps/expval.py
@@ -86,21 +86,12 @@ def run_circuit_and_get_expval(
         instruction_qubits = [qubit.index for qubit in instr[1]]
         active_qubits.extend(instruction_qubits)
 
-
-    # Get the qubits we would like to measure
-    # Data for identities is not stored, need to account for empty terms
-    op_qubits = [term[0][0] for op in ops for term in op.terms if term]
-
-    need_to_activate = set(op_qubits) - set(active_qubits)
+    need_to_activate = set(range(len(qc.qubits))) - set(active_qubits)
     if need_to_activate:
+        # We activate all the qubits for convenience
         for qubit in need_to_activate:
             # Apply the identity
             qc.id(qubit)
-
-    # If there are still no instructions, apply identity to the first qubit
-    # Can happen for an empty circuit when measuring the identity operator
-    if not qc.data:
-        qc.id(qc.qubits[0])
 
     # Convert to zquantum.core.circuit.Circuit
     circuit = Circuit(qc)
--- a/steps/steps_requirements.txt
+++ b/steps/steps_requirements.txt
@@ -1,6 +1,6 @@
-git+https://github.com/zapatacomputing/qe-qiskit#dev
-git+https://github.com/zapatacomputing/qe-qulacs#dev
-git+https://github.com/zapatacomputing/qe-forest#dev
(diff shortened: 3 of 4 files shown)


requirements.txt
```diff
@@ -1,2 +1,2 @@
-pennylane>=0.13.0
+git+https://github.com/PennyLaneAI/pennylane.git@tape_hash
```
antalszava: Temporary, just for the CI to pull the correct branch.

steps/steps_requirements.txt
```diff
@@ -1,10 +1,6 @@
...
-git+https://github.com/zapatacomputing/qe-qiskit#dev
-git+https://github.com/zapatacomputing/qe-qulacs#dev
-git+https://github.com/zapatacomputing/qe-forest#dev
-git+https://github.com/zapatacomputing/z-quantum-core.git#dev
-git+https://github.com/zapatacomputing/qe-openfermion#dev
+git+https://github.com/zapatacomputing/qe-qiskit@dev
+git+https://github.com/zapatacomputing/qe-qulacs@dev
+git+https://github.com/zapatacomputing/qe-forest@dev
+git+https://github.com/zapatacomputing/z-quantum-core.git@dev
+git+https://github.com/zapatacomputing/qe-openfermion@dev
```
josh146, line 5: Is this because these packages defined their own versions of Qiskit etc?

steps/expval.py
```diff
@@ -86,22 +86,13 @@ def run_circuit_and_get_expval(
...
 
-
-    # Get the qubits we would like to measure
-    # Data for identities is not stored, need to account for empty terms
-    op_qubits = [term[0][0] for op in ops for term in op.terms if term]
-
-    need_to_activate = set(op_qubits) - set(active_qubits)
+    need_to_activate = set(range(len(qc.qubits))) - set(active_qubits)
     if need_to_activate:
+        # We activate all the qubits for convenience
```
josh146, line 91: Makes sense, is there an underlying reason for this change? An update to how Orquestra behaves?

tests/test_orquestra_device.py
```diff
@@ -249,14 +249,19 @@ def circuit():
...
-        qnode = qml.QNode(circuit, dev)
-        qnode._construct([], {})
 
-        qasm = dev.serialize_circuit(qnode.circuit)
+        with monkeypatch.context() as m:
+            m.setattr(dev, "execute", lambda *args, **kwargs: np.array([None]))
+            qnode = qml.QNode(circuit, dev)
+            qnode()
+
+        circuit = qnode.qtape.graph if qml.tape_mode_active() else qnode.circuit
```
josh146: I suggest not having `qml.tape_mode_active()` here as it is temporary and will be removed from PL
```suggestion
        circuit = qnode.qtape.graph if hasttr(qnode, "qtape") else qnode.circuit
```

tests/test_orquestra_device.py
```diff
@@ -265,10 +270,14 @@ def circuit():
...
 
-        qnode = qml.QNode(circuit, dev)
-        qnode._construct([], {})
+        with monkeypatch.context() as m:
+            m.setattr(dev, "execute", lambda *args, **kwargs: np.array([None]))
+            qnode = qml.QNode(circuit, dev)
+            qnode()
 
-        qasm = dev.serialize_circuit(qnode.circuit)
+        circuit = qnode.qtape.graph if qml.tape_mode_active() else qnode.circuit
```
josh146:
```suggestion
        circuit = qnode.qtape.graph if hasttr(qnode, "qtape") else qnode.circuit
```

josh146 (approved):
Looks good from my end, just a few questions :)

steps/expval.py
antalszava: Yes, some Orquestra packages were updated (e.g., `z-quantum-core`) and this workaround had to be adjusted to include an identity on all wires.

steps/steps_requirements.txt
antalszava: Yes, rather than pinning here packages explicitly, we would now depend on how they are pinned in the packages. This is required so that if there's a dependency change there, then we wouldn't have to change the requirements in this file.

The other change was adjusting the way how the `dev` branch is used (previously had the incorrect syntax).

PR #20, opened 2021-04-12
Remove analytic keyword argument and v0.15 version bump

josh146:


josh146:
@antalszava, any ideas why the observables are not being decomposed properly? 🤔

pennylane_orquestra/qiskit_device.py
```diff
@@ -46,22 +42,20 @@ class QeQiskitDevice(OrquestraDevice):
     qe_module_name = "qeqiskit.simulator"
     qe_function_name = "QiskitSimulator"
 
-    def __init__(self, wires, shots=10000, backend="qasm_simulator", **kwargs):
+    def __init__(self, wires, shots=1000, backend="qasm_simulator", **kwargs):
```
antalszava: How come we have it lowered here, though for Forest we still have 10k?

pennylane_orquestra/orquestra_device.py
```diff
@@ -118,8 +118,8 @@ class OrquestraDevice(QubitDevice, abc.ABC):
 
     observables = {"PauliX", "PauliY", "PauliZ", "Identity", "Hadamard"}
 
-    def __init__(self, wires, shots=10000, analytic=True, **kwargs):
-        super().__init__(wires=wires, shots=shots, analytic=analytic)
+    def __init__(self, wires, shots=None, **kwargs):
```
antalszava, line 121: Worth noting, that previously we had every device except for IBMQ be initialized as `analytic=True`. With the current state of the PR, we have:

`shots=None`:
* Qulacs

`shots is not None`:
* Qiskit
* Forest

It makes sense having `Qiskit` and `Forest` that way because they both have the ability to simulate HW devices and their state vector simulators can also be run in non-analytic mode (one exception is `statevector_simulator` where we have an explicit warning).

pennylane_orquestra/qiskit_device.py
```diff
@@ -46,22 +42,20 @@ class QeQiskitDevice(OrquestraDevice):
...
         if backend == "qasm_simulator":
-            if kwargs.get("analytic", None):
-                # Raise a warning if the analytic attribute was set to True
```
antalszava, line 52: Note: removed this and the following similar warning just for ease of maintainability

antalszava (approved):
Looks great @josh146 :confetti_ball: Running some e2e tests atm, will get back once they pass well

pennylane_orquestra/cli_actions.py
```diff
@@ -235,7 +235,7 @@ def loop_until_finished(workflow_id, timeout=300):
         tar.extractall()
         tar.close()
 
-        with open("workflow_result.json") as json_file:
+        with open(f"{workflow_id}_workflow_result.json") as json_file:
```
antalszava, line 238: This changed on the Orquestra side (the resulting `json` has the prefix now)

pennylane_orquestra/qiskit_device.py
josh146: Oh, I don't know, maybe it was an accident?

pennylane_orquestra/orquestra_device.py
josh146: Oh! Should this be changed to revert it how it was?

antalszava: I think it's okay to keep it as is, it's just a matter of which one we'd like to have as default. Also, we have warnings for both Qiskit and Forest when `shots is None`. Not sure how we could decide if it was the user passing `shots=None` or that the default argument for `shots` is being used (which is `None`). In both we'd emit a warning :thinking:

pennylane_orquestra/qiskit_device.py
antalszava:
```suggestion
    def __init__(self, wires, shots=10000, backend="qasm_simulator", **kwargs):
```

doc/index.rst
```diff
@@ -72,7 +72,7 @@ base :class:`~.OrquestraDevice` class.
     An option for Orquestra, specifies the resources to be specified for each
     workflow step.
 
-``timeout=300``
+``timeout=600``
```
antalszava, line 75: Observed that an increased timeout is required when doing batched execution

antalszava:
Ran the e2e tests, with the current version it should be :+1:

doc/index.rst
josh146: Thanks! I wonder why

CHANGELOG.md
```diff
@@ -1,19 +1,18 @@
...
+* For compatibility with PennyLane v0.15, the `analytic` keyword argument
+  has been removed from all devices. Analytic expectation values can
+  still be computed by setting `shots=None`.
+  [(#16)](https://github.com/XanaduAI/pennylane-orquestra/pull/16)
 
 ### Contributors
 
 This release contains contributions from (in alphabetical order):
 
+Josh Izaac
```
josh146:
```suggestion
Josh Izaac, Antal Száva.
```

PR #29, opened 2022-08-22
Update docs URL

josh146:


antalszava (approved):
Looks good!

antalszava:
Note: `PennyLane-Orquestra component tests / core-tests (3.7) (pull_request)` is no longer relevant with Python 3.7 support being dropped.

antalszava:
The readthedocs check fail is likely due to the outdated use of Python 3.7. It may require logging in to the readthedocs website and making a manual setup there.

Overriding the check as this is blocking further advances.

PR #34, opened 2022-12-13
Removed decompose_hamiltonian from orquestra device

Jaybsoni:
`decompose_hamiltonian` has since been re-named to `pauli_decompose`!

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
-git+https://github.com/PennyLaneAI/pennylane
+pennylane>=0.28
 pyyaml
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
-pennylane>=0.27
+git+https://github.com/PennyLaneAI/pennylane
 pyyaml
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
-git+https://github.com/PennyLaneAI/pennylane#egg=pennylane
+pennylane>=0.27
 pyyaml
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
-git+https://github.com/PennyLaneAI/pennylane
+pennylane>=0.27
 pyyaml
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
-pennylane>=0.22
+git+https://github.com/PennyLaneAI/pennylane#egg=pennylane
 pyyaml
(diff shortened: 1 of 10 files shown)


Jaybsoni:
@antalszava, 

Note, this PR doesn't attempt to fix the core failing tests which seem to be out of date with the latest rigetti-device. Also the documentation is being fixed in a separate PR. This PR just addresses the unit tests to ensure the plugin works with PennyLane 👍🏼

steps/expval.py
```diff
@@ -22,10 +22,11 @@
 import json
 
 import numpy as np
-from openfermion import IsingOperator, QubitOperator
+# from openfermion import IsingOperator, QubitOperator
```
Jaybsoni:
```suggestion
```

requirements.txt
```diff
@@ -1,2 +1,2 @@
-pennylane>=0.22
+git+https://github.com/PennyLaneAI/pennylane
```
antalszava: Would we like to change to `pennylane>=0.28` in a separate PR or change here and merge this PR after releasing PennyLane?

setup.py
```diff
@@ -18,7 +18,7 @@
 with open("pennylane_orquestra/_version.py") as f:
     version = f.readlines()[-1].split()[-1].strip("\"'")
 
-requirements = ["pyyaml", "pennylane>=0.15"]
+requirements = ["pyyaml", "pennylane>=0.27"]
```
antalszava: Would we need `pennylane>=0.28` here too?

requirements.txt
Jaybsoni: Actually I think we only need Pennylane=0.27. That's when the `decompose_hamiltonian` method was swapped.

setup.py
Jaybsoni: No I think 0.27 is sufficient :)
EDIT: Yes turns out we need 0.28!

requirements.txt
antalszava:
```suggestion
pennylane>=0.27
```

requirements.txt
```diff
@@ -1,2 +1,2 @@
-pennylane>=0.22
+pennylane>=0.27
```
Jaybsoni:
```suggestion
git+https://github.com/PennyLaneAI/pennylane
```

setup.py
Jaybsoni:
```suggestion
requirements = ["pyyaml", "pennylane>=0.28"]
```

requirements.txt
antalszava:
```suggestion
pennylane>=0.28
```

antalszava (approved):

PR #38, opened 2023-07-21
Update rtd config to v2

rashidnhm:
This PR updates the RTD configuration to use the new v2 syntax.


timmysilv (approved):
same comments re: `system_packages` and 3.9 but otherwise lgtm!

PR #44, opened 2023-11-02
sunset the orquestra plugin

timmysilv:
as the title says. I'm matching the format used when sunsetting the SF plugin

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -29,6 +29,11 @@ PennyLane-Orquestra Plugin
     :alt: PyPI - Python Version
     :target: https://pypi.org/project/PennyLane-orquestra
 
+\
+
+    **❗ This plugin will not be supported in newer versions of PennyLane. It is compatible with versions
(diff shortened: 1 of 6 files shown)


timmysilv:
note: the component tests are failing because QVM is trying to run on port 5000, and it's suddenly taken by something? unclear what, and probably not worth debugging. We saw similar issues while debugging some rigetti tests, but the port was taken by the Mac OS Control Center in that case, which isn't happening here (because CI runs on ubuntu machines).

We haven't noticed it in a while because the component tests didn't run with the plugin-test-matrix, so the last passing run was [on Aug 29](https://github.com/PennyLaneAI/pennylane-orquestra/actions/runs/6013510869/job/16311136548)

I am formally requesting we ignore the failure, and merge anyway

trbromley (approved):
Thanks @timmysilv!

As a follow up, can we do a release to v0.33 so users installing this on PyPI get the PL<0.34 constraint?

